PR #846, opened 2026-05-25
feat: upgrade TTS read-aloud to full floating media player

### LINKED ISSUE #815: 💡[Feature]: Multi-lingual Voice Read-Aloud for Articles (Accessibility)
### Feature Description

Description: Enhance accessibility by allowing users to listen to health articles like audiobooks, especially useful for visually impaired users or those on the go. 

Technical Scope:

Integrate expo-speech to convert article text to speech (TTS).
Build a floating or inline media player on ArticleDescriptionScreen.tsx with Play, Pause, Stop, and Speed controls.
Detect the language of the article and automatically select the correct TTS language voice. Why it's major: Significant UX and accessibility improvement without requiring heavy backend changes.

### Record

jpdevhub:
- Replace basic speakSection toggle with Play/Pause/Stop/Speed controls
- Use react-native-tts (existing dependency) per project admin guidance
- Add floating media player bar with dark pill UI
- Track chunk position via refs to support pause/resume/speed change
- Remove unused LottieView import and speechingMode state
closes #815

### DIFF
--- a/frontend/src/screens/article/ArticleScreen.tsx
+++ b/frontend/src/screens/article/ArticleScreen.tsx
@@ -10,7 +10,7 @@ import {
   Dimensions,
   Share,
 } from 'react-native';
-import {useCallback, useEffect, useState} from 'react';
+import React, {useCallback, useEffect, useRef, useState} from 'react';
 import FontAwesome from '@expo/vector-icons/FontAwesome';
 import {ON_PRIMARY_COLOR, PRIMARY_COLOR} from '../../helper/Theme';
 import {SafeAreaView} from 'react-native-safe-area-context';
@@ -35,7 +35,7 @@ import MaterialCommunityIcons from '@expo/vector-icons/MaterialCommunityIcons';
 import {setUserHandle} from '../../store/UserSlice';
 import {FontAwesome5} from '@expo/vector-icons';
 import AutoHeightWebView from '@brown-bear/react-native-autoheight-webview';
-import LottieView from 'lottie-react-native';
+
 import {useGetArticleDetails} from '@/src/hooks/useGetArticleDetail';
 import {useGetArticleContent} from '@/src/hooks/useGetArticleContent';
 import {useGetProfile} from '@/src/hooks/useGetProfile';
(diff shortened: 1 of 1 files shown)


SB2318 (commented):
no objection from myside.

SB2318 (changes requested):
Please fix the Gemini review points.

jpdevhub:
Thank you for the thorough review! I've addressed the feedback and pushed the changes to the branch.

Updates made:

Removed the redundant spakingStarted state entirely to prevent conflicting sources of truth. The pause button logic and cleanup effects now correctly rely on isPlaying and isPaused.
Replaced the magic number 120 with a CHUNK_SIZE constant for clarity.
Updated the status text logic so it properly displays "Stopped" instead of "Loading..." when inactive.
Removed the unused tts-progress listener from the cleanup block.
Cleaned up the useRef calls for consistency.
Added an inline comment acknowledging the minor edge-case with the rewind logic chunk sizing. Since react-native-tts does not expose word-level playback progress, the current chunk-based index fallback remains our best approach.
Let me know if everything looks good for merging!

SB2318 (changes requested):
@jpdevhub why you delete the CartoonModal animation?

Have you checked the visual?

jpdevhub:
Apologies for that! While restructuring the UI to implement the new floating media player, I completely overlooked the TalkBotAnimation component. Since the new player took up visual space at the bottom, I mistakenly assumed it was meant to replace the cartoon visual.

jpdevhub:
I have just pushed a new commit to the PR that restores the LottieView animation block right below the headphone button.

I also took the opportunity to clean up its logic—the animation is now correctly tied to the new isPlaying state instead of the old, redundant spakingStarted state. This means the bot will animate perfectly in sync with the active TTS playback, pausing when paused and hiding when stopped.

SB2318 (approved):
Thanks for your contribution!

SB2318:
## Admin Task


**🔴 High Severity Fix — Player hidden after natural completion**
- In `speakNextChunk`, when `chunkIndex >= words.length` (all audio read), it now calls:
  - `setPlayerVisible(false)` — dismisses the floating player bar
  - `Tts.removeAllListeners('tts-finish')` & `removeAllListeners('tts-error')` — cleans up listeners so no stale callbacks remain
- Previously, only `isPlaying` and `isPaused` were reset, leaving the player bar stuck on screen.

**🟡 Minor Fix — Removed redundant `disabled` prop**
- The `disabled={!isPlaying && !isPaused}` on the pause/resume button was removed. Since the player bar is only rendered when `playerVisible === true`, and `playerVisible` is only set to `true` when playback starts (with `isPlaying = true`), the condition `!isPlaying && !isPaused` is never `true` while the button is visible — so the `disabled` prop was effectively dead code and mildly misleading.

jpdevhub:
Apologies for the issue i will be fixing it shortly.

PR #937, opened 2026-05-30
fix: most viewed articles query never executes due to contradictory enabled condition

### LINKED ISSUE #935: 🐞[Bug]: Most Viewed Articles query never executes due to contradictory enabled condition
### Related Area

Frontend

### Describe the bug

The "Most Viewed Articles" section never loads because the React Query hook is permanently disabled by a logically impossible `enabled` condition.

File:
`useGetMostViewedArticle.ts`

Current implementation:

```ts
enabled:
  !!isConnected &&
  !!(!userId && others) &&
  !!(!user_id && !others),
```

This condition can never evaluate to `true`.

* If `others` is `true`, then `!others` becomes `false`, causing the last condition to fail.
* If `others` is `false`, then the middle condition fails.

As a result, the query never runs and article data is never fetched.

The issue is visible in the Activity Overview section where the "Most Viewed Articles" component renders but receives no data.

### Expected behavior

The query should execute when valid user information is available and the application is connected.

The "Most Viewed Articles" section should fetch and display article data instead of remaining empty.

A possible fix would be:

```ts
enabled: !!isConnected && (others ? !!userId : !!user_id),
```

or any equivalent condition that correctly reflects the intended behavior.

### Add ScreenShots

N/A

The issue was identified through code inspection. The query's enabled condition is logically contradictory and prevents execution in all scenarios.

Chrome

### Record

Akshita-2307:
## Summary

The `enabled` condition in `useGetAuthorMostViewedArticles` was:

```ts
enabled: !!isConnected && !!(!userId && others) && !!(!user_id && !others)
```

This requires `others` to be simultaneously `true` (sub-condition 2) AND `false` (sub-condition 3) — a logical impossibility. As a result, the query is permanently disabled and the "Most Viewed Articles" section in Activity Overview always shows empty.

**Fix:** Changed to:
```ts
enabled: !!isConnected && ((others && !!userId) || (!others && !!user_id))
```

This correctly enables the query when `isConnected` is true AND either:
- `others=true` with a valid `userId`, OR
- `others=false` with a valid `user_id`

Fixes #935

### DIFF
--- a/frontend/src/hooks/useGetMostViewedArticle.ts
+++ b/frontend/src/hooks/useGetMostViewedArticle.ts
@@ -27,6 +27,6 @@ export const useGetAuthorMostViewedArticles = ({
       return response.data as ArticleData[];
     },
 
-    enabled: !!isConnected && !!(!userId && others) && !!(!user_id && !others),
+    enabled: !!isConnected && ((others && !!userId) || (!others && !!user_id)),
   });
 };


Akshita-2307:
Hi maintainers! 👋 This PR fixes a bug where the 'Most Viewed Articles' query never executes because the enabled condition in React Query is contradictory (it checks the same variable it's supposed to set). Without this fix, the most-viewed articles section always shows a loading/empty state.

This is a small, focused 1-line change that's been rebased on latest main. Would appreciate a review when you get a chance. Thank you!

PR #972, opened 2026-06-01
fix: resolve impossible enabled condition in most viewed articles query

### LINKED ISSUE #935: 🐞[Bug]: Most Viewed Articles query never executes due to contradictory enabled condition

Krishnx21:
# Description

## Bug

The `useGetAuthorMostViewedArticles` query was permanently disabled due to a contradictory `enabled` condition:

```ts
enabled:
  !!isConnected &&
  !!(!userId && others) &&
  !!(!user_id && !others)
```

This condition required `others` to be both `true` and `false` simultaneously, making it impossible for React Query to execute the request.

As a result, the **Most Viewed Articles** section never received data and could not render correctly.

## Fix

Updated the query `enabled` condition to match the actual UI rendering path:

```ts
enabled: !!isConnected && !!others && !!userId
```

## Why this approach?

After tracing the code flow:

* `useGetAuthorMostViewedArticles` is only used inside `ActivityOverview`.
* The **Most Viewed Articles** section is rendered only when `others === true`.
* The section is not displayed on the current user's own profile.

The query now executes only when:

* The application is connected.
* Another user's profile is being viewed.
* A valid `userId` is available.

This fixes the bug while avoiding unnecessary requests for hidden UI paths.

## Additional Improvements

* Replaced index-based list keys with stable article identifiers (`pb_recordId || _id`).
* Added fallback image handling for articles without images.
* Updated image access to use `item.imageUtils?.[0]` for consistent null-safe handling.
* Added safe author ID extraction to support both string and object-based `authorId` values.
* Added an empty state when no most-viewed articles are available.

## Scope

The unrelated `useUploadPodcast.ts` modification was reverted to keep this PR focused on the Most Viewed Articles issue.

## Testing

* Verified the original `enabled` condition was logically impossible.
* Verified the query executes when viewing another user's profile.
* Verified the query does not execute when `userId` is unavailable.
* Verified the query does not execute while offline.
* Verified no unnecessary request is made for hidden own-profile paths.

Fixes #935


Krishnx21:
Hi @SB2318 

I investigated the failing CI checks.

This PR only modifies the Most Viewed Articles query hook. Neither `src/screens/auth/OtpScreen.tsx` nor `src/helper/APIUtils.ts` are part of this PR's changes.

The reported TypeScript failures are coming from `OtpScreen.tsx`, and the lint failures appear to originate from files outside the scope of this change. My modification is limited to the query `enabled` condition and query key for the Most Viewed Articles hook.

I have also addressed the PR Governance check by adding the appropriate label.

Please let me know if you'd like me to further reduce the PR scope or investigate any specific failure.

SB2318 (changes requested):
Overall, great catch and solid fix for the root issue in `useGetAuthorMostViewedArticles`. The explanation and updated logic make sense and clearly resolve the feature not rendering.

A couple of points to address before approval:

1. **Scope creep in `useUploadPodcast.ts`** — the change to use the provided `article_id` instead of explicitly sending `null` appears unrelated to this PR’s goal. Please either revert it or explicitly justify it in the PR description with expected impact and testing details.
2. **Consistency in null handling** — consider using `item.imageUtils?.[0]` at the call site for consistency with the null-safe pattern already used elsewhere. Current behavior is fine because the helper handles undefined, but aligning usage improves readability.

Positive notes:

* Good fix for the impossible React Query `enabled` condition.
* Nice improvement using stable list keys instead of indexes.
* Helper extraction (`getArticleImageSource`, `getArticleAuthorId`) improves maintainability and readability.

Requesting changes mainly because of the unrelated functional update; once clarified or separated, this looks in good shape.

Krishnx21:
Thanks for the review.

I addressed both requested points:

* Reverted the unrelated `useUploadPodcast.ts` change so this PR stays focused on the Most Viewed Articles bug.
* Updated the image handling for consistency by passing `item.imageUtils?.[0]` at the call site.

The PR diff is now limited to the Most Viewed Articles hook and the related `ActivityOverview` changes.

I've pushed the updates and updated the PR description accordingly.

Thanks again for the feedback and re-review.

SB2318 (approved):
Thanks for your contribution!

Issue #1012, opened 2026-06-02
[Feature] Create Reusable Glossary Tooltip Component for Medical Definitions


SB2318:
## Overview

Create a reusable tooltip/popover component that displays glossary definitions when users interact with highlighted medical terms in articles.

The component should be generic and reusable across the application so it can support future glossary and educational content features.

## Problem

Health articles often contain medical terminology that may be unfamiliar to readers. Currently, users must leave the article to search for definitions elsewhere, disrupting the reading experience.

## Proposed Solution

Build a reusable glossary tooltip component that:

* Displays a definition when a user clicks, taps, or hovers on a glossary term.
* Supports both desktop and mobile interactions.
* Appears as a tooltip or popover anchored to the selected term.
* Can be reused throughout the application.
* Supports accessibility requirements.

### Component API Example

```tsx
<GlossaryTooltip
  term="Hypertension"
  definition="A condition in which blood pressure remains consistently high."
>
  Hypertension
</GlossaryTooltip>
```

### Functional Requirements

* Show definition on hover (desktop).
* Show definition on tap/click (mobile and desktop).
* Close when clicking outside.
* Proper positioning near the selected term.
* Support long definitions with scrolling if necessary.
* Prevent viewport overflow.
* Support keyboard navigation.
* Support screen readers.

### Accessibility

* ARIA-compliant tooltip/popover.
* Focus management.
* Keyboard support (`Enter`, `Space`, `Esc`).

### Acceptance Criteria
* [ ] Share Screenshot
* [ ] Reusable glossary tooltip component created.
* [ ] Works on desktop and mobile.
* [ ] Accessible via keyboard and screen readers.
* [ ] Handles viewport boundaries correctly.
* [ ] Supports configurable content and styling.
* [ ] Covered by component tests.

## Mockup

```text
The patient was diagnosed with [Hypertension]

                    ┌──────────────────────────────┐
                    │ Hypertension                 │
                    │ High blood pressure that     │
                    │ persists over time.          │
                    └──────────────────────────────┘
```

### Screenshot / Design Reference

[...]

PR #1031, opened 2026-06-02
feat: add GlossaryTooltip component for medical term definitions (#1012)

InukaWijerathna:
Fixes #1012

## What's added

A `GlossaryTooltip` component (`web/src/components/ui/glossary-tooltip.tsx`) that wraps medical terms in articles so readers can see a definition without leaving the page.

Built on the Radix UI `Popover` primitive which is already in the project's dependencies, and follows the same patterns as the other shadcn/ui components in `web/src/components/ui/`.

## How it works

On desktop, hovering over a term opens the tooltip. There's a small 150ms delay before it closes so you can move your cursor onto the tooltip itself without it disappearing on you. On mobile, tap to open, tap again or tap outside to close.

Keyboard users can tab to the term and use Enter/Space to toggle, Escape to close. Long definitions scroll inside the tooltip rather than overflowing.

## Usage

`	sx
import { GlossaryTooltip } from "@/components/ui/glossary-tooltip"

<p>
  Patients with{" "}
  <GlossaryTooltip
    term="Hypertension"
    definition="A condition in which blood pressure remains consistently high."
  >
    hypertension
  </GlossaryTooltip>{" "}
  should monitor their sodium intake.
</p>
`

## Testing checklist

- [ ] Hover over term — tooltip opens and stays open when cursor moves onto the content
- [ ] Move cursor away — tooltip closes after the brief delay
- [ ] Tap on mobile — toggles open/closed
- [ ] Tab to focus, Enter to open, Escape to close
- [ ] Paste in a long definition — content scrolls inside the tooltip
- [ ] Reposition near a viewport edge — Radix repositions it automatically
- [ ] Check in dark mode

### DIFF
--- a/web/src/components/ui/glossary-tooltip.tsx
+++ b/web/src/components/ui/glossary-tooltip.tsx
@@ -1,0 +1,80 @@
+"use client"
+
+import * as React from "react"
+import { Popover as PopoverPrimitive } from "radix-ui"
+
+import { cn } from "@/lib/utils"
+
+interface GlossaryTooltipProps {
+  term: string
+  definition: string
+  children: React.ReactNode
+  className?: string
+}
+
+function GlossaryTooltip({ term, definition, children, className }: GlossaryTooltipProps) {
+  const [open, setOpen] = React.useState(false)
+  const closeTimerRef = React.useRef<ReturnType<typeof setTimeout> | null>(null)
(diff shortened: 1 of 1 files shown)


SB2318 (approved):
Thanks for your contribution!

PR #1049, opened 2026-06-03
[Refactor] Remove unused user_id parameter from useGetAuthorMostViewedArticles (#991)

### LINKED ISSUE #991: [Refactor] Remove unused user_id parameter from useGetAuthorMostViewedArticles
### Followed By PR #972 

You can check **Gemini Guidance**

## Description
The `useGetAuthorMostViewedArticles` hook still accepts `user_id`, but this value is never used when the query is active.

Keeping this parameter creates unnecessary complexity and makes the hook API harder to understand for future maintainers.

## Impact
- Unclear hook contract  
- Dead/unnecessary parameter  
- Reduced maintainability  

## Expected Fix
- Remove `user_id` from the hook signature  
- Update all existing usages accordingly  
- Verify query behavior remains unchanged after cleanup  

## Labels
`refactor` `cleanup` `frontend` `medium-priority`

Xenon010101:
﻿Removes the unused user_id parameter from useGetAuthorMostViewedArticles. The hook is only rendered when others is true (viewing someone else's profile), where userId is always provided. The broken enabled condition (always false due to contradictory &&) is replaced with !!isConnected && !!userId.

**Changes:**
- Removed user_id from hook signature and query logic
- Simplified URL to always use userId
- Fixed enabled condition
- Updated the single call site in ActivityOverview.tsx

Closes #991

### DIFF
--- a/frontend/src/components/ActivityOverview.tsx
+++ b/frontend/src/components/ActivityOverview.tsx
@@ -150,7 +150,6 @@ const ActivityOverview = ({
   // GET MOST VIEWED ARTICLE
   const {data: article, isLoading: isArticleLoading} =
     useGetAuthorMostViewedArticles({
-      user_id: user_id,
       userId: userId,
       others: others,
       isConnected: isConnected,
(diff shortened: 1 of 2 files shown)


SB2318:
@Xenon010101 You are not assigned to the issue

Issue #1046, opened 2026-06-03
🐞[Bug]: Launching Soon modal is unreachable because comingSoonModal is never set to true


Krishnx21:
### Related Area

Frontend

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

The homepage contains a "Launching Soon" modal controlled by the comingSoonModal state.

const [comingSoonModal, setComingSoonModal] = useState(false);

The modal is rendered conditionally:

<div className={`modal-overlay${comingSoonModal ? " active" : ""}`}>

However, there is no code path that calls:

setComingSoonModal(true);

As a result, the modal can never become visible to users, making the feature inaccessible.

### Expected behavior

The modal should open when a user interacts with the intended trigger element (button, card, feature link, etc.).

Instead, the state remains false throughout the component lifecycle, preventing the modal from ever being displayed.

### What browsers are you seeing the problem on?

Chrome

### Record

- [x] I have read the Contributing Guidelines
- [ ] I'm a GSSOC'24 contributor
- [x] I'm a GSSOC'26 contributor
- [ ] I'm a IEEE IGDTUW contributor
- [x] I want to work on this issue

Krishnx21:
/ASSIGN/GSSOC 2026

DakshataDalvi:
Hi, I'd like to work on this issue. Please assign it to me.

PR #1104, opened 2026-06-04
fix: make Launching Soon modal reachable by adding trigger action

Krishnx21:
## Description

This PR fixes the issue where the "Launching Soon" modal could never be displayed because `comingSoonModal` was initialized as `false` and there was no code path that updated it to `true`.

### Changes Made

* Added a trigger to call `setComingSoonModal(true)` when users interact with the intended UI element.
* Ensured the modal becomes visible when the associated feature/button/card is clicked.
* Preserved the existing modal close behavior.
* Verified that the modal opens and closes correctly without affecting other homepage functionality.

### Problem

Previously:

```js
const [comingSoonModal, setComingSoonModal] = useState(false);
```

The modal was conditionally rendered based on `comingSoonModal`, but the state was never updated to `true`, making the modal unreachable for users.

### Solution

Implemented the missing trigger so the modal state is updated appropriately and the feature is now accessible.

### Related Issue

Closes #1046 

### Testing

* Verified the trigger opens the modal.
* Verified the modal closes correctly.
* Checked that no existing homepage functionality was affected.


SB2318 (commented):
Thanks for adding the missing trigger for the "Launching Soon" modal.

However, I noticed that the existing "UHealth Admin" action appears to have been changed from a direct Google Play Store link to a button that only opens the modal.

Could you clarify whether this behavior change was intentional and approved?

From the previous implementation, users could directly navigate to the UHealth Admin Play Store page. With this PR, that navigation path seems to be removed entirely and replaced by the modal. Since the issue is about making the modal reachable, this introduces a broader functional change than the scope described in the PR.

If the app is genuinely not available yet, replacing the link may be appropriate. Otherwise, we should consider preserving the existing navigation behavior or documenting this change explicitly in the PR description.

Also, please confirm whether the removed Play Store link was intentionally deprecated.

Krishnx21:
@SB2318 thank you for reviewing i have updated it .

SB2318 (approved):
Thanks for your contribution!

SB2318:
#  Post-Approval Review Findings

### 🔴 Blocking: Unresolved Merge Conflicts
`web/src/app/page.tsx` still contains merge conflict markers (`<<<<<<<`, `=======`, `>>>>>>>`).

This leaves the file in an invalid state and will prevent successful builds and merges. Please resolve the conflicts and ensure only the intended final code remains.

---

### 🔴 Blocking: Unrelated Submodule Modification
This PR includes an update/addition to the `ieee-submodules/IEEE-s-Mindful-Devs-Bootcamp` submodule.

The stated purpose of this PR is to make the **"Launching Soon"** modal reachable. The submodule change appears unrelated to that goal and significantly increases the review scope.

Please remove the submodule modification from this PR and submit it separately if it is required.

---

### 🟡 Needs Clarification: Login Layout Changes
The PR removes `PageWrapper` and `Section` from both login pages and replaces them with generic `div` containers.

These changes appear unrelated to the modal accessibility fix. If they are intentional, please explain the reason and confirm there are no styling, layout, or accessibility regressions. Otherwise, consider reverting them to keep the PR focused on a single concern.

---

### 🟢 Minor: Import Cleanup
There appears to be a duplicate `ScrollToTop` import in `web/src/app/page.tsx`.

This may be a byproduct of the unresolved merge conflict. After resolving conflicts, please verify imports are cleaned up and deduplicated.

---

## ✅ What's Good

- The modal accessibility issue appears to be addressed through the new callback flow.
- The added `aria-label` improves accessibility for screen reader users.
- Using `Pick<HeroAndDownloadProps, "onShowComingSoon">` keeps child component props focused and type-safe.
- The solution follows the existing component hierarchy without introducing unnecessary complexity.

---

## Summary

The primary feature implementation looks reasonable and appears to solve the original issue. However, the presence of unresolved merge conflicts and unrelated submodule modifications makes the PR unmergeable in its current state.

Please address the blocking issues and provide clarification for the login page layout changes before resubmission.

Krishnx21:
@SB2318 THANKS FOR REVIEWING 
Addressed review feedback:
- Removed unrelated submodule change from this PR.
- Resolved merge-conflict residue and duplicate imports in `web/src/app/page.tsx`.
- Restored login pages to use `Section` and `PageWrapper`.
- Preserved the UHealth Admin Play Store link while adding a separate Launch Status modal trigger.

SB2318 (approved):

PR #1122, opened 2026-06-05
fix: title and description alignment for long content on PodcastDetail screen

Hasini2706:
#### PR Description

This PR resolves Issue #1095 by improving how long podcast titles and descriptions are displayed on the PodcastDetail screen.

### Changes Made

* Fixed layout behavior for long podcast titles to ensure proper wrapping and prevent clipping.
* Improved description rendering for long content with better spacing and readability.
* Added expandable "Read More / Read Less" functionality for lengthy descriptions.
* Improved accessibility by removing rigid text sizing constraints that could cause overlap when system font scaling is enabled.
* Enhanced responsiveness for different screen sizes and orientations.
* Improved ScrollView content handling to prevent controls and metadata from being pushed off-screen.
* Added test coverage for long title rendering and expandable description behavior.

#### Type of Change

* [x] Bug fix (change which fixes an issue)
* [ ] New feature (change which adds functionality)
* [ ] Documentation update

#### Select your work-area

* [x] Frontend
* [ ] Backend
* [ ] Documentation
* [ ] Others

#### Related Issue

Fixes the PodcastDetail screen layout issue for long content.

#### Add your Work Example

📷 Screenshot/GIF demonstrating long titles and descriptions rendering correctly will be attached.

#### Fixes (mention the issue number which this fixes)

Closes #1095

#### Checklist

* [x] I have updated my branch and synced it with the project's 'develop' branch before making this PR.
* [x] I have optimized the file changes.
* [ ] I have added a snapshot of my work example.
* [x] I have made a PR to the project's develop branch.

#### Undertaking

* [x] My code follows the style guidelines of this project.

* [x] I have performed a self-review of my code.

* [x] I have commented my code, particularly in hard-to-understand areas.

* [x] I have made corresponding changes to the documentation.

* [x] I have checked for plagiarism and assure its authenticity.

* [x] I have read and followed the code of conduct for this repository. I understand that violation of this undertaking may have legal consequences.

* [x] I Agree

PR #1144, opened 2026-06-05
fix: resolve stale state in pull-to-refresh and tie spinner to actual data load (#1042)

### LINKED ISSUE #1042: [Bug] Pull-to-refresh reads stale React state — data refetch silently skips on non-first pages
### Description

while profiling the feed reload behavior, i found a React state closure bug in onRefresh that causes pull-to-refresh to not actually reload data when the user has scrolled past page 1.

### The Problem

HomeScreen.tsx lines 433-453: setPage(1) schedules an async state update but page === 1 on line 439 reads the STALE value. if user is on page 3, page is still 3 when checked so refetch() is never called. setRefreshing(false) also fires synchronously so the spinner disappears instantly while data loads in background.

user sees spinner flash and disappear with no visible update, then data silently changes seconds later.

### Proposed Fix

- remove the if (page === 1) guard — always call refetch() after setPage(1)
- or use useEffect to trigger refetch when page transitions to 1
- move setRefreshing(false) into the querys onSettled callback

pull-to-refresh is broken for anyone who scrolls past page 1 — app feels unresponsive and stale.

for labels i think gssoc:approved + level:intermediate + type:bug + quality:clean fits here.

vedant7007:
Closes #1042

pull-to-refresh was broken when the user scrolled past page 1. two bugs:

**bug 1 — stale page closure:** setPage(1) is async but page === 1 was read synchronously on the same tick. if user was on page 3, the condition failed and refetch() never ran.

**bug 2 — premature spinner:** setRefreshing(false) fired synchronously so the spinner disappeared instantly while data was still loading in the background.

what i fixed:

stale page (approach: remove guard, always refetch):
- removed the if (page === 1) guard entirely — refetch() now always fires
- when page is already 1, refetch re-fetches correctly. when page > 1, setPage(1) triggers a queryKey change that auto-fetches page 1
- simpler than adding a flag + useEffect, works correctly in both cases

premature spinner (approach: useEffect on isFetching):
- removed synchronous setRefreshing(false) from inside onRefresh
- added useEffect watching [isFetching, refreshing] — spinner only dismisses when React Querys actual fetch completes
- chose this over await refetch() because refetch() resolves for the current queryKey, not the page-1 auto-fetch after state change. isFetching correctly waits for ALL fetching to finish

bonus fixes:
- allArticlesRef.current = [] on refresh — clears stale accumulated articles from pages 1-N
- lastCategoryFilteredCountRef.current = -1 on refresh — resets sparse-category auto-pagination guard
- removed unnecessary debug console.log in refresh handler

1 file changed. typecheck passes.

for labels i think gssoc:approved + level:intermediate + type:bug + quality:clean fits here. happy to make changes if needed!

### DIFF
--- a/frontend/src/screens/HomeScreen.tsx
+++ b/frontend/src/screens/HomeScreen.tsx
@@ -444,19 +444,16 @@ const HomeScreen = ({navigation}: HomeScreenProps) => {
 
 
   const onRefresh = () => {
-    if (__DEV__) console.log('is connected', isConnected);
     if (isConnected) {
       setRefreshing(true);
-      // Reset pagination state on full pull-to-refresh
+      allArticlesRef.current = [];
+      lastCategoryFilteredCountRef.current = -1;
       setPage(1);
-      if (page === 1) {
-        refetch();
-      }
+      refetch();
       if (!isGuest) {
         refetchUser();
         refetchUnreadCount();
       }
-      setRefreshing(false);
     } else {
       Snackbar.show({
         text: 'Please check your network connection',
@@ -464,6 +461,13 @@ const HomeScreen = ({navigation}: HomeScreenProps) => {
       });
     }
   };
+
+  useEffect(() => {
+    if (refreshing && !isFetching) {
+      setRefreshing(false);
+    }
+  }, [isFetching, refreshing]);
+
   const handleSearch = (textInput: string) => {
     //console.log('Search Input', textInput);
     if (textInput === '' || articleData === undefined) {


vedant7007:
Hi maintainers,

Regarding the 3 failing checks — none are related to this PR's changes:

1. **Unit Tests** — `GlossaryBottomSheet.test.tsx` fails with `No component given to styled()` (Tamagui setup issue). All 77 actual tests pass. This is a recently added test that appears to be broken upstream.

2. **Knip Dead Code Check** — The workflow references `frontend/pnpm-lock.yaml` for cache, but that file was removed in a recent commit. Fails before any analysis runs.

3. **PR Governance** — Requires at least one label, which only maintainers can apply. Could you add `bug` when convenient?

PR #1122 has the same 3 failures for the same reasons. Typecheck, lint, and Expo Doctor all pass on this PR.

Happy to make any changes to the actual code if needed!

SB2318 (approved):
Thanks for your contribution!

Issue #1067, opened 2026-06-04
[BUG]: Article Bottom Sheet Intermittently Fails to Open from Three-Dot Menu


SB2318:
## Description

When users tap the three-dot menu on an article card, the article bottom sheet does not consistently appear. The issue is intermittent—sometimes the bottom sheet opens as expected, while other times nothing happens despite the tap being registered.

## Steps to Reproduce

1. Navigate to the Articles screen.
2. Tap the three-dot menu on an article card.
3. Observe that the bottom sheet may not appear.
4. Repeat the action on the same or different articles.

## Current Behavior

* Bottom sheet opens inconsistently.
* Users may need to tap multiple times before the bottom sheet appears.
* In some cases, no action is triggered after tapping the three-dot menu.

## Expected Behavior

* The article bottom sheet should open immediately and reliably every time the three-dot menu is tapped.

## Possible Areas to Investigate

* Bottom sheet ref initialization and lifecycle.
* Race conditions between state updates and bottom sheet presentation.
* Touch event handling on the three-dot menu button.
* Re-renders causing the bottom sheet state to reset.
* Modal/bottom sheet stacking and visibility issues.

## Acceptance Criteria

* [ ] Article bottom sheet opens consistently on every tap of the three-dot menu.
* [ ] No multiple taps are required.
* [ ] Behavior is reliable across all article lists and screens.
* [ ] No regressions are introduced to existing bottom sheet functionality.
* [ ] Verified on Android and iOS devices.

<img width="720" height="1600" alt="Image" src="https://github.com/user-attachments/assets/56c2f468-b04c-43f7-841d-2e99ca378b12" />

<img width="576" height="1280" alt="Image" src="https://github.com/user-attachments/assets/acad61b1-61df-4a5c-8d04-b91d7bd6fef8" />

SarthakKharche:
Hi @SB2318  👋

I’d like to work on this article bottom sheet reliability issue.

I reviewed the bug description and agree that inconsistent bottom sheet behavior can significantly impact usability, especially when users expect immediate access to article actions from the three-dot menu.

Planned investigation and implementation:

* Analyze the three-dot menu press handler and bottom sheet trigger flow
* Verify bottom sheet ref initialization and lifecycle management
* Check for race conditions between state updates and bottom sheet presentation
* Review touch event handling to ensure taps are always registered and processed
* Investigate potential re-render issues that may reset the selected article or bottom sheet state
[...]

PR #1190, opened 2026-06-06
fix(mobile): resolve bottom sheet toggle and expand menu tap target

aayushprsingh:
This PR resolves issue #1067 by fixing the intermittent bottom sheet opening state tracking and expanding the three-dot menu tap area with hitSlop.

Closes #1067


SB2318 (changes requested):
@aayushprsingh Please create PR for the Main branch, and please delete all unrelated files.

Please also fix the bot comments in next PR.

PR #1234, opened 2026-06-07
fix: remove offline podcast from UI after deletion

### LINKED ISSUE #1231: 🐞[Bug]: Deleted offline podcasts remain visible until the screen reloads
## Related Area

Frontend

## Is there an existing issue for this?

- [x] I searched open and closed issues and pull requests for offline podcast deletion/list refresh behavior.

## Describe the bug

Removing a downloaded podcast from the Offline Podcasts screen updates persistent download storage, but the deleted podcast card remains visible until the screen is reopened.

`OfflinePodcastList` calls `deleteFromDownloads(item)` from the card's download action. On success it only shows a snackbar; it never removes the matching item from the local `podcasts` state or reloads that state. `deleteFromDownloads` correctly writes the filtered list to storage, so the visible list and persisted list become inconsistent for the rest of the current screen session.

## Location

- `frontend/src/screens/OfflinePodcastList.tsx:83-97` (`downLoadAudio` callback)
- `frontend/src/helper/Utils.ts:478-503` (`deleteFromDownloads`)

## Expected behavior

After a successful deletion, the removed podcast should disappear from the list immediately. If deletion fails, the existing card should remain and the failure message should be shown.

## Suggested approach

After `deleteFromDownloads(item)` returns success, update `podcasts` by filtering out `item._id` (or reload the persisted list). Add a focused screen test covering successful and failed deletion so the UI cannot drift from download storage again.

Requesting this be considered for GSSoC 2026; I would be happy to work on the focused fix and regression test.

codematser69:
Summary

Fixes the issue where deleted offline podcasts remain visible on the Offline Podcasts screen until the page is reloaded.

Changes Made

- Updated the local "podcasts" state immediately after a successful deletion.
- Removed the deleted podcast from the displayed list using "setPodcasts()".
- Kept existing success and failure snackbar notifications unchanged.

Problem

Previously, "deleteFromDownloads(item)" removed the podcast from persistent storage, but the UI state was not updated. As a result, the deleted podcast card remained visible until the screen was refreshed or reopened.

Solution

After a successful deletion, the podcast is now removed from the local state using:

setPodcasts(prev =>
  prev.filter(podcast => podcast._id !== item._id)
);

This ensures the UI stays synchronized with storage and provides immediate visual feedback to users.

Issue

Closes #1231

### DIFF
--- a/frontend/src/screens/OfflinePodcastList.tsx
+++ b/frontend/src/screens/OfflinePodcastList.tsx
@@ -85,11 +85,15 @@ export default function OfflinePodcastList({
           const res = await deleteFromDownloads(item);
 
           if (res) {
-            Snackbar.show({
-              text: 'Podcast has been removed from offline',
-              duration: Snackbar.LENGTH_SHORT,
-            });
-          } else {
+  setPodcasts(prev =>
+    prev.filter(podcast => podcast._id !== item._id)
+  );
+
+  Snackbar.show({
+    text: 'Podcast has been removed from offline',
+    duration: Snackbar.LENGTH_SHORT,
+  });
+} else {
             Snackbar.show({
               text: 'Failed to removed podcast from offline',
               duration: Snackbar.LENGTH_SHORT,


SB2318:
@codematser69, you are not assigned to the issue.

Issue #1432, opened 2026-06-11
🐞[Bug]: Broken Image and Misalignment in "UltimateHealth App" Carousel


Aditya7005446231:
### Related Area

Frontend

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

In the "Screenshots" section of the landing page, under the UltimateHealth App card, the third mobile screenshot is failing to load. This broken image icon disrupts the UI and causes a noticeable misalignment within the image carousel.

### Expected behavior

All images in the carousel should render successfully.
The cards should maintain a uniform height and consistent alignment (likely using align-items: stretch or a fixed aspect ratio) even if an image fails to load, preventing the layout from breaking.
A proper fallback image or empty state should be displayed if the source URL is invalid.

### Add ScreenShots

<img width="1108" height="796" alt="Image" src="https://github.com/user-attachments/assets/153b60f0-0b57-4442-8202-ec1b18226263" />

### What browsers are you seeing the problem on?

Chrome

### Record

- [x] I have read the Contributing Guidelines
- [ ] I'm a GSSOC'24 contributor
- [ ] I'm a GSSOC'26 contributor
- [ ] I'm a IEEE IGDTUW contributor
- [x] I want to work on this issue

RajD2910:
hey @SB2318 , I would like to work on this issue if it gets available , i am contributing under GSSOC'26.
Thankyou

nityanara:
Hi @SB2318, 

I would like to work on this issue under GSSoC '26. Kindly assign this issue to me.

dyuthialva:
Hello @SB2318    I’d like to work on this issue under GSSoC 2026.

Anuj1804:
Hi, I would like to work on this issue under GSSoC. Kindly assign it to me.

PR #1504, opened 2026-06-13
#1432 fix: correct broken image path in UltimateHealth App carousel

patilrutuja23:
## Description

This PR fixes a broken image in the UltimateHealth App carousel.

### Root Cause

The carousel referenced:

`/assets/UltimateHealth-about.jpeg`

while the actual file present in `public/assets` is:

`/assets/ultimate-health-about.jpeg`

This filename mismatch caused image loading failures and generated 400 errors from Next.js image optimization.

### Changes Made

* Updated the image path to use the correct filename:

  * `/assets/ultimate-health-about.jpeg`

### Verification

* Confirmed the image loads correctly in the carousel.
* Confirmed browser console no longer shows 400 errors for this image request.
* Verified carousel rendering after the fix.

### DIFF
--- a/web/src/app/page.tsx
+++ b/web/src/app/page.tsx
@@ -27,7 +27,7 @@ const userScreenshots = [
   { src: "/assets/podcast-recording.jpeg", caption: "Podcast Recorder" },
   { src: "/assets/podcast-upload.jpeg", caption: "Podcast Upload" },
   { src: "/assets/notificaion-screen.jpeg", caption: "Notification" },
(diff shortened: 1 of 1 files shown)


patilrutuja23:
I found this issue while testing the project locally and verified that the broken image was caused by an incorrect image path (`UltimateHealth-about.jpeg` vs `ultimate-health-about.jpeg`). The PR contains only that path correction. If this issue is reserved for the original assignee, I completely understand and can close the PR.

Issue #1514, opened 2026-06-13
🐞[Bug]: ArticleDescriptionScreen Category Selection Fails Due to Reference-Based Comparison


Simarpreet2005:
### Related Area

Frontend

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

The category/tag selection logic in ArticleDescriptionScreen.tsx relies on object reference comparison when determining whether a category is selected.

Because categories may be recreated from API responses or state updates, object references can differ even when the underlying category represents the same item. As a result:

Previously selected categories may not appear selected.
Users may be unable to deselect categories correctly.
Selection state becomes inconsistent after data refreshes or re-renders.

The current implementation appears to compare entire objects rather than stable identifiers such as id or _id.

### Expected behavior

Category selection should be based on unique identifiers (id or _id) instead of object references.

Users should be able to:

Select categories consistently.
Deselect categories correctly.
Retain accurate selection state after re-renders or data refreshes.

### Add ScreenShots

N/A

### Record

- [x] I have read the Contributing Guidelines
- [ ] I'm a GSSOC'24 contributor
- [x] I'm a GSSOC'26 contributor
- [ ] I'm a IEEE IGDTUW contributor
- [x] I want to work on this issue

PR #1516, opened 2026-06-13
fix: resolve EmailInputModal TypeScript error

Simarpreet2005:
#### PR Description
This PR fixes the category selection logic in ArticleDescriptionScreen.tsx.

Previously, selected categories were checked using object reference comparison. When category objects were recreated after API responses or component re-renders, previously selected categories could fail to appear selected and users could experience inconsistent selection behavior.

#### Related Issue

Status: Open.
#1514


SB2318 (changes requested):
Why so unrelated changes?

Please provide the context of every file changes.

Issue #1088, opened 2026-06-04
## Bug: Podcast Recording Timer Desyncs When App Is Backgrounded


Pcmhacker-piro:
## Bug: Podcast Recording Timer Desyncs When App Is Backgrounded

---

### Description

When a user records a podcast episode using the `PodcastRecorderScreen`, the recording timer (elapsed time display) desyncs if the app is sent to the background and then reopened. The timer continues incrementing at the wrong rate or jumps forward by the duration the app was backgrounded, causing the displayed recording duration to not match the actual audio file length. This leads to incorrect timestamps and a confusing user experience.

---

### Steps to Reproduce

1. Navigate to the Podcast Recorder screen
2. Tap "Start Recording"
3. Observe the timer counting up normally
4. Press the home button or switch to another app (background the app)
5. Wait 30-60 seconds
6. Return to the UltimateHealth app
7. Observe that the timer has jumped forward by significantly more than the actual elapsed recording time, or is incrementing at an irregular pace
8. Stop the recording — the displayed duration does not match the actual audio file length

---

### Expected Behavior

- The recording timer should use a reliable time source (e.g., `Date.now()` offset or the audio recorder's actual tracked duration) rather than `setInterval` which is paused/throttled in the background
- On app foreground, the timer should re-sync to the actual recording duration from the audio recorder API
- The final displayed duration should exactly match the audio file's actual length
- No visual glitches or jumps in the timer display

---

### Implementation Hints

The issue is likely in how the recording timer is implemented. A common pattern that causes this:

```typescript
// Buggy pattern — interval-based, pauses in background
const [elapsed, setElapsed] = useState(0);
useEffect(() => {
  const interval = setInterval(() => {
    setElapsed(prev => prev + 1); // Breaks in background
  }, 1000);
  return () => clearInterval(interval);
}, []);
```

Fix by using start timestamp offset and checking actual elapsed time:

```typescript
// Fixed pattern — uses start time offset
const [elapsed, setElapsed] = useState(0);
const startTimeRef = useRef<number>(0);

useEffect(() => {
  startTimeRef.current = Date.now();
  const interval = setInterval(() => {
    const actualElapsed = Math.floor((Date.now() - startTimeRef.current) / 1000);
    setElapsed(actualElapsed);
  }, 1000);
  return () => clearInterval(interval);
}, []);
```

Also handle app state changes using `AppState` listener:

```typescript
import { AppState, AppStateStatus } from 'react-native';

[...]

PR #1530, opened 2026-06-13
fix(podcast): resolve recording timer desync when app is backgrounded    issue (#1088)

dinesh9997:
# Fix: Keep Recording Timer Synchronized with Native Audio Duration

## Overview

This PR fixes an issue where the recording timer could become inaccurate after the application was backgrounded and later brought back to the foreground.

Previously, the timer relied on JavaScript-based interval calculations:

```typescript
Date.now() - recordStartTimeRef.current
```

Because `setInterval` is throttled or suspended when the app is in the background, the displayed recording duration could drift from the actual duration of the recorded audio file.

---

## Root Cause

When the application entered the background:

- The JavaScript thread was paused or throttled.
- `setInterval` callbacks stopped running reliably.
- The timer no longer tracked elapsed recording time accurately.
- On returning to the foreground, the displayed duration could lag behind or jump forward unexpectedly.
- The UI timer and generated audio duration could become inconsistent.

---

## Solution

The recording timer has been refactored to use the native recorder's tracked duration as the single source of truth.

#### 1. Native Duration Polling

Configured `useAudioRecorderState` with a polling interval of **500ms** to continuously retrieve the recorder's native duration.

```typescript
useAudioRecorderState(recorder, 500);
```

#### 2. Dynamic Timer Synchronization

Added a synchronization effect that listens to `durationMillis` updates and adjusts the timer offset accordingly.

```typescript
useEffect(() => {
  // Sync displayed timer with native recorder duration
}, [recorderState.durationMillis]);
```

This prevents timer drift while recording is active.

#### 3. Foreground Re-Synchronization

Implemented an `AppState` listener to immediately re-sync elapsed time when the application returns to the foreground.

```typescript
AppState.addEventListener("change", (state) => {
  if (state === "active") {
    // Recalculate timer using native duration
  }
});
```

This ensures accurate timing even after extended periods in the background.

#### 4. TypeScript Compatibility Improvements

Updated typings to avoid environment-specific compilation issues.

```typescript
const intervalRef = useRef<any>(null);

(state: any) => {
  // AppState handler
};
```

---

## Files Modified

```text
frontend/src/screens/PodcastRecorder.tsx
```

---

### Type Checking

```bash
npm run type-check
```

Result:

```text
✓ No TypeScript errors
✓ Successful compilation
```

### Functional Testing

Verified the following scenarios:

- Start recording and keep the app in the foreground.
- Background the application during recording.
- Return to the foreground after several seconds/minutes.
- Continue recording.
- Stop recording and compare displayed duration with generated audio length.

Results:

```text
✓ Timer remains synchronized with native recorder duration
✓ No noticeable drift after background/foreground transitions
✓ Displayed duration matches actual recorded audio duration
✓ Consistent behavior across app lifecycle changes
```

---

## Impact

- Eliminates timer drift caused by JavaScript thread suspension.
- Keeps UI duration aligned with actual audio recording duration.
- Improves recording reliability during app lifecycle transitions.
- Provides a more accurate and consistent user experience.

---

## Closes

**#1088**

### DIFF
--- a/frontend/src/screens/PodcastRecorder.tsx
+++ b/frontend/src/screens/PodcastRecorder.tsx
@@ -1,5 +1,5 @@
 import React, {useCallback, useEffect, useRef, useState} from 'react';
-import {StyleSheet, Alert} from 'react-native';
+import {StyleSheet, Alert, AppState} from 'react-native';
 
 import {PodcastRecorderScreenProps} from '../type';
 import RNFS from 'react-native-fs';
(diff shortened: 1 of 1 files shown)


dinesh9997:
Thank you for the feedback. I have addressed the review comments by improving type safety and adding the requested code documentation.

Changes made:

* Replaced `any` types with appropriate TypeScript types (`number | null` for the timer reference and `AppStateStatus` for the AppState listener).
* Added comments explaining the purpose of `durationMillisRef` and its role in preventing stale closures.
* Added clarification for the fallback logic used when native duration data is not yet available during the initial recording phase.

The latest commit includes these updates. Please let me know if any further changes are required.

SB2318 (approved):

PR #1541, opened 2026-06-14
fix: set guest mode on cold start when no stored token exists

### LINKED ISSUE #1536: 🐞[Bug]: Guest mode is never set on cold start without a stored token, causing spurious "Session Expired" alerts for new/logged-out users
### Related Area

Frontend

### Describe the bug

In `AppContent.tsx`, `checkToken()` only dispatches `setGuestMode(false)` when a token is found in secure storage. If no token exists (first launch, fresh install, or after a manual logout that doesn't go through the 401 path), `isGuest` remains at its initial value of `false` (UserSlice.ts) while `user_token` becomes `null`.

Because many data-fetching hooks gate on `enabled: !isGuest` (e.g. useGetUserDetails, useGetPlaylists, useGetUserAllImprovements, useGetNotificationPreferences), these queries fire with no Authorization header attached (the axios request interceptor deletes the header when no token is present). The backend responds 401, which is caught by `handleError` in setupAxiosInterceptor.ts. That handler then:
- dispatches setUserToken(''), setUserId(''), setUserHandle(''), setGuestMode(true)
- clears secure storage
- shows a "Your session has expired. You are now browsing as a guest." toast/alert

Reproduction: fresh install → open app with network connected → observe Session Expired toast despite never having authenticated.

### Expected behavior

This means a user who has never logged in (or who is correctly anonymous) sees a "session expired" message on the very first app open, purely as a side effect of unauthenticated queries firing before guest mode is correctly established.

### Add ScreenShots

_No response_

_No response_

### Record

saurabhhhcodes:
checkToken() only dispatched setGuestMode(false) when a token was found in secure storage, but never set guest mode to true on first launch, fresh install, or after a manual logout.

Added an else branch that dispatches setGuestMode(true) when no token is present, fixing the spurious 'Session Expired' alerts for new/logged-out users.

Closes #1536

### DIFF
--- a/frontend/src/components/AppContent.tsx
+++ b/frontend/src/components/AppContent.tsx
@@ -52,6 +52,8 @@ export default function AppContent() {
       dispatch(setUserToken(token));
       if (token) {
         dispatch(setGuestMode(false));
+      } else {
+        dispatch(setGuestMode(true));
       }
       initDeepLinking(navigationRef.current, tokenRes?.isValid || false);
     
--- a/frontend/src/components/EmailVerificationModal.tsx
+++ b/frontend/src/components/EmailVerificationModal.tsx
@@ -1,0 +1,130 @@
+import React from 'react';
+import {
+  Modal,
+  View,
+  Text,
+  TouchableOpacity,
+  StyleSheet,
+  ActivityIndicator,
+} from 'react-native';
+
+interface EmailVerificationModalProps {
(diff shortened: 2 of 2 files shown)


saurabhhhcodes:
The fix was already applied upstream with the refactored checkToken + useEffect pattern. Closing this PR.

PR #1542, opened 2026-06-14
fix: prevent duplicate mark-as-read requests on rapid scroll

### LINKED ISSUE #1537: 🐞[Bug]: Rapid onScroll events trigger duplicate "mark article as read" requests due to async-only completion guard
### Related Area

Frontend

### Describe the bug

In `ArticleScreen.tsx`, `handleReadStatusUpdate` (called via `runOnJS` from `useAnimatedScrollHandler`'s `onScroll`) guards the `updateReadEvent` mutation with `!readEventSave`, a React state value that is only flipped to `true` inside the mutation's `onSuccess` callback.

`onScroll` fires many times per second while the user is at (or oscillating around) the bottom of the article. Each of those calls re-evaluates `handleReadStatusUpdate` against the *same* stale `readEventSave === false` value until the in-flight mutation resolves and triggers a re-render. Because `useMutation` does not dedupe concurrent `mutate()` calls by itself, every scroll-tick in that window fires a fresh POST to `UPDATE_READ_EVENT` for the same article.

Reproduction: open a short article, scroll to the bottom and let the scroll settle with small jitter (or fling repeatedly) — multiple `update-read-event-status` requests for the same `article_id` are fired before the first response returns.

### Expected behavior

Notably, a `readEventFiredRef` (`useRef(false)`) is declared and reset on articleId change but is never read or set anywhere outside its declaration — it appears to be a leftover from an intended synchronous in-flight guard that was never wired up.

### Add ScreenShots

_No response_

_No response_

### Record

saurabhhhcodes:
onScroll fires many times per second when the user is at the bottom of the article. Each call checked `readEventSave` from `useState`, which captured a stale snapshot of the guard value due to closure over the React state.

Replaced `useState` with `useRef` so the onScroll closure always reads the latest value, preventing duplicate `updateReadEvent` mutations.

Closes #1537

### DIFF
--- a/frontend/src/components/AppContent.tsx
+++ b/frontend/src/components/AppContent.tsx
@@ -52,6 +52,8 @@ export default function AppContent() {
       dispatch(setUserToken(token));
       if (token) {
         dispatch(setGuestMode(false));
+      } else {
+        dispatch(setGuestMode(true));
       }
       initDeepLinking(navigationRef.current, tokenRes?.isValid || false);
     
--- a/frontend/src/components/EmailVerificationModal.tsx
+++ b/frontend/src/components/EmailVerificationModal.tsx
@@ -1,0 +1,130 @@
+import React from 'react';
+import {
+  Modal,
+  View,
+  Text,
+  TouchableOpacity,
+  StyleSheet,
+  ActivityIndicator,
+} from 'react-native';
+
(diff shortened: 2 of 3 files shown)


saurabhhhcodes:
Closing - will re-evaluate with a cleaner fix.

PR #1866, opened 2026-06-23
fix: remove duplicate Reading History row from ProfileScreen (Fixes #1865)

### LINKED ISSUE #1865: ProfileScreen.tsx duplicates "Reading History" — once as a row and again as a History tab
### Related Area

Frontend

### Describe the bug

# `ProfileScreen.tsx` duplicates "Reading History" — once as a row and again as a `History` tab

## Description

`ProfileScreen.tsx` renders a **Reading History row** (a `TouchableOpacity` that navigates to `ReadingHistoryScreen`) **and** a **History tab** inside the tab bar — both showing the same reading history data. This is a UX duplication that confuses users about which entry to use.

The `History` tab inside the profile renders a paginated `FlatList` of `ReadHistoryArticle` items directly inside the profile `ScrollView`. The row shortcut navigates away to a separate `ReadingHistoryScreen`.

## File

`frontend/src/screens/ProfileScreen.tsx`

## Suggested Fix

Pick one approach:
1. **Keep only the `History` tab** and remove the quick-access row — the tab is already paginated and embedded inline.
2. **Keep only the row shortcut** and remove the `History` tab from the tab bar — to keep the Profile screen lean and delegate to a dedicated screen.

The `History` tab is the more complete implementation (with pagination, error state, and retry), so removing the row shortcut (option 1) is recommended.

### Expected behavior

The Profile screen should expose reading history once, through the inline History tab (which already has pagination, error/retry state, and load-more). The quick-access "Reading History" row that navigates away to a separate ReadingHistoryScreen is redundant and should be removed, since it duplicates the same data via a different path and breaks the tab-based navigation pattern already established on the screen.

### Add ScreenShots

_No response_

_No response_

### Record

Tomeshwari-02:
## Description

`ProfileScreen.tsx` exposed reading history in two places at the same time:

1. A `TouchableOpacity` row labelled "Reading History" that navigated away to the separate `ReadingHistoryScreen`.
2. A `History` tab in the inline tab bar that rendered a paginated `FlatList` directly inside the profile `ScrollView`.

Both pointed to the same data, confusing users about which entry to use and breaking the tab-based navigation pattern already established on the screen.

The `History` tab is the more complete implementation — it has pagination, error state with retry, and a load-more footer — so the redundant row shortcut has been removed.

## Changes

`frontend/src/screens/ProfileScreen.tsx`

- Removed the `TouchableOpacity` "Reading History" row and its `onPress` navigation to `ReadingHistoryScreen`
- Removed the `Ionicon` import (`@expo/vector-icons/Ionicons`) which was only used by that row
- Removed the now-dead `historyRow` and `historyRowText` style entries

Fixes #1865

## Checklist
- [x] My code follows the project's style guidelines
- [x] I have verified this change manually
- [x] No new dependencies introduced

### DIFF
--- a/frontend/src/screens/ProfileScreen.tsx
+++ b/frontend/src/screens/ProfileScreen.tsx
@@ -18,7 +18,6 @@ import {useGetProfile} from '../hooks/useGetProfile';
 import {useUpdateViewCount} from '../hooks/useUpdateViewCount';
 import {useGetReadHistory, ReadHistoryArticle} from '../hooks/useGetReadHistory';
 import { NoArticleState } from '../components/EmptyStates';
-import Ionicon from '@expo/vector-icons/Ionicons';
 
 const ProfileScreen = ({navigation}: ProfileScreenProps) => {
   const theme = useTheme();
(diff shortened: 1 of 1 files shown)


Tomeshwari-02:

SB2318 (commented):
Thanks for your contribution! But I cannot accept the changes this time, As the **ProfileScreen** Architecture has changed.

PR #1922, opened 2026-06-25
fix: resolve external user profile routes

### LINKED ISSUE #1913: 🐞[Bug]: UserProfileScreen ignores authorId and opens the signed-in user profile
### Related Area

Frontend / Navigation / Profiles

### Describe the bug

`frontend/src/screens/UserProfileScreen.tsx` reads `authorId` from the route, but calculates the query target as:

```ts
const actualAuthorId = routeUserId || user_id;
```

This ignores the commonly supplied `authorId` parameter. Callers such as `SocialScreen` and `NotificationScreen` navigate with `{ authorId: ... }`, so `routeUserId` is undefined and the screen falls back to the currently signed-in user's ID.

### Steps to reproduce

1. Sign in to the app.
2. Open another user's followers/following list or tap a user-follow notification.
3. Tap a different user's profile.
4. Observe that the signed-in user's profile is requested instead.

### Expected behavior

The screen should resolve the target consistently from `authorId`, `userId`, or handle before falling back to the signed-in user. All callers and `RootStackParamList` should use one documented parameter shape.

### Impact

- Other-user profile links can open the wrong account.
- Follow/unfollow actions may target the wrong user.
- Notification and social-list navigation appear broken.

### Suggested fix

Normalize the route parameters once and use the resolved external user ID in `useGetAuthorProfile`, statistics, follow actions, and focus refresh logic. Add navigation tests for both ID- and handle-based routes.

Tomeshwari-02:
## Summary

- normalize `authorId`, `userId`, `author_handle`, and `userHandle` route inputs
- avoid falling back to the signed-in user when an external handle is provided
- use the resolved profile owner for statistics, follow mutations, and follow notifications
- remove the profile-route debug log

## Root cause

`UserProfileScreen` read `authorId` from navigation params but only used `userId` when selecting the profile query target. Most callers pass `authorId`, so those routes fell back to the signed-in user's ID.

## Validation

- `git diff --check`
- reviewed all `UserProfileScreen` navigation call sites for ID and handle parameter shapes
- dependency installation was attempted twice but Yarn 1.22.22 fails during package fetching on Windows with `EISDIR: illegal operation on a directory, read`

Fixes #1913


Tomeshwari-02:
Addressed in bca8903. Added a dedicated profile-route resolver with regression tests covering string `authorId`, object `authorId`, legacy `userId`, both handle parameter forms, and the current-user fallback.

Tomeshwari-02:

SB2318 (commented):
Thanks for your contribution @Tomeshwari-02 But, as of now I can't accept the PR, as Our 
**UserProfileScreen** problem already resolved, and Please study from the latest branch.

PR #1923, opened 2026-06-25
fix: target follow notifications correctly

### LINKED ISSUE #1916: 🐞[Bug]: SocialScreen emits follow notifications with a stale or empty recipient ID
### Related Area

Frontend / Social / Real-time notifications

### Describe the bug

In `frontend/src/screens/SocialScreen.tsx`, the follow button stores the selected user in state and then uses that state inside the mutation callback:

```ts
setUserId(follower._id);
followMutate(follower._id, {
  onSuccess: () => {
    socket.emit('notification', {
      type: 'userFollow',
      userId: userid,
    });
  },
});
```

React state updates are asynchronous. The `userid` captured by `onSuccess` can still be the previous value, and on the first follow action it is initially an empty string.

### Steps to reproduce

1. Open a followers/following list.
2. Follow a user for the first time.
3. Inspect the emitted `notification` socket payload.
4. The payload can contain `userId: ''` or the ID of a previously selected user.

### Expected behavior

The follow notification must always be emitted to the user who was just followed.

### Impact

- The intended recipient may receive no notification.
- A different user may receive the notification after consecutive follow actions.
- Follow state and notification state become inconsistent.

### Suggested fix

Use `follower._id` directly inside `onSuccess` instead of routing the recipient through component state. Add a test covering two consecutive follow operations and assert each emitted recipient ID.

Tomeshwari-02:
## Summary

- remove the intermediate recipient state from `SocialScreen`
- emit follow notifications directly to the user whose follow mutation succeeded
- remove the now-unused `useState` import

## Root cause

The previous implementation called `setUserId(follower._id)` and then read `userid` from the mutation callback. Because React state updates are asynchronous, the callback could emit an empty or previously selected recipient ID.

## Validation

- `git diff --check`
- confirmed no remaining `userid` or `setUserId` references in `SocialScreen`
- dependency installation remains blocked by the repository's Yarn/Windows `EISDIR` fetch failure

Fixes #1916


Tomeshwari-02:
Addressed in 74c3905. Extracted the follow-notification emission into a focused helper and added tests verifying that the successful follow target's ID is used as the notification recipient, including the no-socket path.

SB2318 (approved):

PR #1925, opened 2026-06-25
fix: merge notification pages safely

### LINKED ISSUE #1915: 🐞[Bug]: Notification pagination can overwrite pages because it appends from stale state
### Related Area

Frontend / Notifications / Pagination

### Describe the bug

`frontend/src/screens/NotificationScreen.tsx` appends later pages using the `notificationsData` value captured by the effect closure:

```ts
const oldNotif = notificationsData ?? [];
setNotificationsData([...oldNotif, ...notificationsRes.notifications]);
```

The effect depends on `notificationsRes` and `page`, but not `notificationsData`. If page requests resolve close together, a later response can append to an older array and discard notifications already added by another response. Refetched pages can also introduce duplicate IDs.

### Steps to reproduce

1. Open the notification screen with multiple pages of data.
2. Scroll quickly enough to trigger successive page requests, or refetch while another page is loading.
3. Let responses complete out of order.
4. Observe missing or duplicated notifications.

### Expected behavior

Each successful page should merge into the latest list without losing existing items, and duplicate notification IDs should be ignored.

### Suggested fix

Use a functional state update such as `setNotificationsData(previous => mergeById(previous, incoming))`, or migrate this endpoint to `useInfiniteQuery`. Add a test where page responses resolve out of order.

Tomeshwari-02:
## Summary

- append notification pages with a functional state update
- merge pages by notification ID to prevent duplicate rows
- preserve already loaded pages when responses complete close together

## Root cause

Later pages were appended from the `notificationsData` array captured by the effect closure. A response could therefore merge into an older snapshot and overwrite notifications added by another response.

## Validation

- `git diff --check`
- verified page one still replaces the list while later pages merge into the latest state

Fixes #1915

### DIFF
--- a/frontend/src/helper/notificationUtils.test.ts
+++ b/frontend/src/helper/notificationUtils.test.ts
@@ -1,0 +1,40 @@
+import {mergeNotificationsById} from './notificationUtils';
+
+const notification = (id: string, title = id) =>
+  ({_id: id, title}) as any;
+
+describe('mergeNotificationsById', () => {
+  it('returns an empty list when both pages are empty', () => {
+    expect(mergeNotificationsById([], [])).toEqual([]);
+  });
+
+  it('preserves the order of unique notifications', () => {
+    expect(
+      mergeNotificationsById(
+        [notification('1'), notification('2')],
+        [notification('3')],
+      ).map(item => item._id),
+    ).toEqual(['1', '2', '3']);
+  });
+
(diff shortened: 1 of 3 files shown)


Tomeshwari-02:
Addressed in 1b9471e. Moved the merge logic into a focused utility and added tests for empty arrays, unique-page ordering, duplicate IDs with updated values, and duplicates within an incoming page.

SB2318 (approved):

PR #1928, opened 2026-06-25
fix: encode podcast search parameters

### LINKED ISSUE #1919: 🐞[Bug]: Podcast search builds malformed URLs for reserved characters
### Related Area

Frontend / Podcast search

### Describe the bug

`frontend/src/hooks/useGetSearchPodcasts.ts` interpolates raw user input directly into the URL:

```ts
axios.get(`${SEARCH_PODCAST}?q=${q}&page=${page}`)
```

Reserved URL characters in the search term are interpreted as query-string syntax. For example, `heart & health` produces an extra parameter after `&`, `#` starts a fragment, and `+` may be decoded as a space.

`useGetAuthorProfile.ts` uses the same manual query-string construction pattern for handles.

### Steps to reproduce

1. Open podcast search.
2. Search for a term containing `&`, `#`, `+`, `%`, or `?`.
3. Inspect the outgoing request.
4. The `q` value is truncated or changed by URL parsing, producing incorrect results.

### Expected behavior

The complete search text should be transmitted as one query parameter regardless of reserved characters.

### Suggested fix

Pass query values through Axios `params`:

```ts
axios.get(SEARCH_PODCAST, {params: {q, page}})
```

Alternatively, use `URLSearchParams` or `encodeURIComponent`. Add tests for spaces and reserved characters.

Tomeshwari-02:
## Summary

- pass podcast search input through Axios `params`
- preserve reserved characters as part of the `q` value
- keep pagination as a separately encoded query parameter

## Root cause

The search hook interpolated raw user input into the URL. Characters such as `&`, `#`, `+`, `%`, and `?` could be interpreted as query-string syntax instead of search text.

## Validation

- `git diff --check`
- verified Axios receives `q` and `page` as separate parameter values

Fixes #1919

### DIFF
--- a/frontend/src/hooks/useGetSearchPodcasts.ts
+++ b/frontend/src/hooks/useGetSearchPodcasts.ts
@@ -15,7 +15,12 @@ AxiosError
   return useQuery({
     queryKey: ['search-podcasts', page, q],
     queryFn: async () => {
-      const res = await axios.get(`${SEARCH_PODCAST}?q=${q}&page=${page}`);
+      const res = await axios.get(SEARCH_PODCAST, {
+        params: {
+          q,
+          page,
+        },
+      });
 
       return res.data as SearchResponse;
     },


Tomeshwari-02:

SB2318 (commented):
Sorry @Tomeshwari-02, but this doesn't follow the contribution guidelines.

If our GitHub bot has already determined that the assignment requirements were not met, the issue should be reopened instead of continuing work on it. In that case, no further implementation or PR should be submitted until the issue is reassigned.

The correct workflow is:

1. Submit your PR for the issue currently assigned to you.
2. Once that's done, comment on the new issue you'd like to work on.
3. Wait for our GitHub bot to assign the issue to you.
4. Only after the assignment is confirmed should you start working and create a PR.

This process ensures that issues are assigned fairly and prevents multiple contributors from working on the same issue simultaneously.

PR #1929, opened 2026-06-25
fix: validate stored reading progress

### LINKED ISSUE #1920: 🐞[Bug]: Corrupted reading-progress storage causes an unhandled parse failure
### Related Area

Frontend / Article reading progress / Local storage

### Describe the bug

`frontend/src/services/ReadingProgressService.ts` parses persisted progress without validating or handling malformed data:

```ts
const raw = await MMKVUtils.getItem(key);
return raw ? JSON.parse(raw) : null;
```

If the stored value is truncated, belongs to an older incompatible format, or is otherwise invalid JSON, `JSON.parse` rejects. `ArticleDescriptionScreen` calls `getProgress(...).then(...)` without a rejection handler, creating an unhandled promise rejection during screen initialization.

### Steps to reproduce

1. Store an invalid value for a `reading_progress_<articleId>` key, such as `{broken`.
2. Open the corresponding article description screen.
3. Observe the rejected promise/error instead of a safe fallback.

### Expected behavior

Corrupted reading-progress data should be treated as missing data. The invalid key should be removed or replaced, and the article screen should continue normally.

### Suggested fix

Wrap parsing in `try/catch`, validate the parsed object and numeric ranges, clear malformed entries, and return `null`. Add tests for invalid JSON, missing properties, and out-of-range percentages.

Tomeshwari-02:
## Summary

- replace the invalid generic `MMKVUtils` import with the existing AsyncStorage dependency
- validate article ID, percentage range, and timestamp before returning stored progress
- remove malformed or incompatible progress records
- clamp saved percentages to the supported 0-100 range
- add service tests for valid, malformed, mismatched, clamped, and cleared data

## Root cause

Stored progress was passed directly through `JSON.parse` without error handling or runtime validation. Corrupted data rejected the promise during article initialization. The service also referenced a generic `MMKVUtils` API that the current helper does not export.

## Validation

- `git diff --check`
- added `ReadingProgressService.test.ts` with five focused cases
- tests could not be executed locally because dependency installation fails during Yarn package fetching on Windows with `EISDIR: illegal operation on a directory, read`

Fixes #1920

### DIFF
--- a/frontend/src/helper/MMKVUtils.ts
+++ b/frontend/src/helper/MMKVUtils.ts
@@ -21,9 +21,12 @@ export type PodcastDownloadRecord = Omit<SerializedPodcastDownloadRecord, 'downl
 
 const PODCAST_STORAGE_KEY = 'DOWNLOAD_PODCAST_DATA';
 const PODCAST_CACHE_ID = 'podcast_cache';
+const READING_PROGRESS_CACHE_ID = 'reading_progress_cache';
 
 let podcastMMKV: MMKVStorageLike | null = null;
 let hasAttemptedInitialization = false;
+let readingProgressMMKV: MMKVStorageLike | null = null;
+let hasAttemptedReadingProgressInitialization = false;
 
 const initializeMMKV = (): MMKVStorageLike | null => {
   if (hasAttemptedInitialization) {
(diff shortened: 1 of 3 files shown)


SB2318 (changes requested):
Do not use async storage; replace it with `mmkv`.
Use the dedicated `MMKVUtils.ts` files for your task.
Make sure you have integrated your service with the main app functionality.

SB2318 (approved):

PR #1931, opened 2026-06-25
fix: complete logout when offline

### LINKED ISSUE #1921: 🐞[Bug]: Users cannot log out locally when the logout API is unavailable
### Related Area

Frontend / Authentication / Logout

### Describe the bug

`frontend/src/screens/auth/LogoutScreen.tsx` clears credentials only inside the server mutation's `onSuccess` callback:

```ts
logout({}, {
  onSuccess: async () => {
    await clearStorage();
    dispatch(resetUserState());
    navigation.reset(...);
  },
  onError: () => {
    Alert.alert('Logout Failed', ...);
  },
});
```

When the device is offline, the API is unavailable, or the server returns an error, local credentials remain stored and the user cannot end the local session.

### Steps to reproduce

1. Sign in normally.
2. Disable network connectivity.
3. Open Logout and confirm.
4. The request fails and an alert is shown.
5. Restart the app; the local session is still present.

### Expected behavior

A user must always be able to log out locally. Server-side token revocation can be best-effort, queued, or retried, but network failure should not prevent local credential removal and navigation to the login screen.

### Impact

- Users on unreliable networks cannot sign out.
- A shared or lost device may retain access because the server is unreachable.
- Logout behavior depends on backend availability.

### Suggested fix

Clear secure/local storage, Redux auth state, socket state, and user-scoped caches in a `finally` path. Attempt remote revocation when possible, and document whether failed revocations are retried later. Add an offline logout test.

Tomeshwari-02:
## Summary

- centralize local logout cleanup and navigation
- execute local logout after both successful and failed remote revocation attempts
- distinguish confirmed server logout from offline/local-only logout feedback
- remove unused authentication imports

## Root cause

Local credentials and Redux state were cleared only from the mutation's success callback. A network or server error therefore prevented users from ending the local session.

## Validation

- `git diff --check`
- reviewed success, HTTP-error, and network-error paths

Fixes #1921

### DIFF
--- a/frontend/src/screens/__tests__/LogoutScreen.local.test.tsx
+++ b/frontend/src/screens/__tests__/LogoutScreen.local.test.tsx
@@ -1,0 +1,126 @@
+import React from 'react';
+import {Alert} from 'react-native';
+import {fireEvent, render, waitFor} from '@testing-library/react-native';
+import LogoutScreen from '../auth/LogoutScreen';
+
+const mockDispatch = jest.fn();
+const mockReset = jest.fn();
+const mockClearStorage = jest.fn(() => Promise.resolve());
+const mockLogout = jest.fn();
+
+jest.mock('react-redux', () => ({
+  useDispatch: () => mockDispatch,
+}));
+
+jest.mock('../../hooks/useUserLogout', () => ({
+  useUserLogout: () => ({
+    mutate: mockLogout,
+  }),
+}));
+
+jest.mock('../../helper/Utils', () => ({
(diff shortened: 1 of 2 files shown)


Tomeshwari-02:
Addressed in 9724474. Added `LogoutScreen` tests for successful server logout, server-error fallback, and offline/network-error fallback. Each case verifies local storage cleanup, user-state reset, navigation, and the appropriate user-facing alert.

SB2318 (approved):

PR #2006, opened 2026-06-28
fix: encode podcast share URLs

### LINKED ISSUE #1996: [Bug] Podcast share links break when audio URLs contain reserved characters
### Problem
Podcast share links are assembled by interpolating `audioUrl` directly into the query string. If the audio URL contains its own `?`, `&`, spaces, signed-token params, or other reserved characters, the generated share endpoint receives a malformed `audioUrl` value.

### Affected code
- `frontend/src/components/PodcastCard.tsx`
- `frontend/src/screens/OfflinePodcastDetails.tsx`

Both build URLs in this shape:

```ts
https://uhsocial.in/api/share/podcast?trackId=${id}&audioUrl=${audioUrl}
```

### Steps to reproduce
1. Open a podcast whose audio URL contains query parameters, for example `audio.mp3?token=abc&expires=123`.
2. Tap the share action from the podcast card or offline podcast detail screen.
3. Inspect the generated share URL.

### Expected behavior
The full audio URL should be preserved as one encoded `audioUrl` query parameter.

### Actual behavior
Reserved characters in the nested audio URL can be interpreted as parameters of the share endpoint, truncating or corrupting the shared audio URL.

### Suggested fix
Build the URL with `URLSearchParams` or `encodeURIComponent` for each query value.

### Acceptance criteria
- `trackId` and `audioUrl` are encoded safely.
- Podcast share links work with audio URLs that contain query strings.
- Both card and offline detail share flows use the same safe helper.

Tomeshwari-02:
## Summary
- added a shared podcast share URL builder that encodes query values safely
- updated podcast card and offline podcast detail sharing to use the helper

## Why
Podcast audio URLs can contain their own query strings or reserved characters. Encoding the nested URL prevents the share endpoint parameters from being split or corrupted.

Fixes #1996

## Validation
- `git diff --check`
- `yarn type-check` was attempted, but the current base branch has unrelated syntax errors in `App.tsx`, `src/hooks/useDeletePocketbaseRecord.ts`, and `src/screens/article/ArticleScreen.tsx`.


SB2318 (changes requested):
Thanks for the contribution! The URL encoding fix using `generatePodcastShareUrl` is a solid improvement, and the accompanying tests are appreciated.

However, I have a few concerns that need to be addressed before we can merge this PR:

1. **`WeeklyChart` refactor**

   * The custom implementation appears to regress accessibility compared to the previous solution. Please ensure the chart exposes meaningful accessibility information for assistive technologies.
   * The current fixed sizing (`barWidth`, `spacing`, etc.) may not scale well across different screen sizes. Please verify and, if necessary, make the layout responsive.

2. **Removal of `ambient-modules.d.ts`**

   * Please confirm that removing these declarations doesn't introduce new TypeScript issues. A clean `yarn type-check` after resolving unrelated errors would help validate this change.

3. **`PROD_URL` fallback**

   * The current fallback (`10.0.2.2`) is Android emulator-specific. Please consider a platform-aware fallback or rely on environment-specific configuration to avoid issues for iOS developers.

There are also a few minor cleanup items (`id={undefined as never}`, `as any` casts, and some unrelated type additions) that would be good to revisit, though they are not blockers by themselves.

Please address the above points and update the PR. Once they're resolved, I'd be happy to take another look.

SB2318:
@Tomeshwari-02 any update?

SB2318:
Sorry, @Tomeshwari-02, cannot accept the contribution! You have not responded yet. Please create a new PR if you want to contribute all the suggested fixes.

PR #2008, opened 2026-06-28
fix: preserve filtered notification preferences

### LINKED ISSUE #1998: [Bug] Notification preferences Clear All removes hidden filtered selections
### Problem
On the notification preferences screen, the `Clear All` action clears every selected topic, including topics that are currently hidden by the search filter. This can remove saved preferences the user cannot see at the time they tap the button.

### Affected code
- `frontend/src/screens/NotificationPreferencesScreen.tsx`

Current behavior:

```ts
onPress={() => setSelectedIds([])}
```

### Steps to reproduce
1. Open notification preferences with several topics already selected.
2. Search for a term that shows only a subset of topics.
3. Tap `Clear All` while the list is filtered.
4. Save preferences.

### Expected behavior
When search is active, the action should either:
- clear only visible filtered topics, or
- clearly warn that all selected topics, including hidden ones, will be removed.

### Actual behavior
All selected IDs are cleared, including hidden selections outside the filtered view.

### Suggested fix
Rename/split the action depending on search state:
- `Clear visible` when a search filter is active;
- `Clear all` only when no filter is active, or after confirmation.

### Acceptance criteria
- Hidden selected preferences are not removed unexpectedly.
- The button label and behavior match the current filter state.
- Add a regression test for filtered clear behavior.

Tomeshwari-02:
## Summary
- changed notification preference clearing to preserve hidden selections while search is active
- updated the bulk action label to `Clear Visible` when filtered

## Why
Users should not lose selected topics they cannot currently see in a filtered list.

Fixes #1998

## Validation
- `git diff --check`

### DIFF
--- a/frontend/App.tsx
+++ b/frontend/App.tsx
@@ -7,17 +7,17 @@ import { logger } from './src/services/monitoring/logger';
 import AppContent from './src/components/AppContent';
 
 function App() {
-const [queryClient] = useState(
-() =>
-new QueryClient({
-defaultOptions: {
-queries: {
-retry: 2,
-staleTime: 1000 * 60,
-},
-},
-})
-);
+  const [queryClient] = useState(
+    () =>
+      new QueryClient({
+        defaultOptions: {
+          queries: {
+            retry: 2,
+            staleTime: 1000 * 60,
+          },
+        },
+      }),
+  );
 
   useEffect(() => {
     let isMounted = true;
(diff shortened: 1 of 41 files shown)


SB2318 (commented):
Sorry @Tomeshwari-02, but this doesn't follow the contribution guidelines.

If our GitHub bot has already determined that the assignment requirements were not met, the issue should be reopened instead of continuing work on it. In that case, no further implementation or PR should be submitted until the issue is reassigned.

The correct workflow is:

1. Submit your PR for the issue currently assigned to you.
2. Once that's done, comment on the new issue you'd like to work on.
3. Wait for our GitHub bot to assign the issue to you.
4. Only after the assignment is confirmed should you start working and create a PR.

This process ensures that issues are assigned fairly and prevents multiple contributors from working on the same issue simultaneously.

PR #2009, opened 2026-06-28
fix: preserve missing notification preferences

### LINKED ISSUE #1999: [Bug] Notification preferences save can drop selected topics missing from category response
### Problem
The notification preferences save payload is rebuilt by filtering the currently loaded category list. If the saved `selectedIds` contain IDs that are not present in the current `categories` response, those IDs are silently dropped on save.

### Affected code
- `frontend/src/screens/NotificationPreferencesScreen.tsx`

Current save logic:

```ts
const payload = (categories ?? []).filter(cat =>
  selectedIds.includes(cat._id),
);
```

### Why this matters
A user can lose existing preferences if:
- the category API response is incomplete;
- categories are paginated in the future;
- a previously selected category is renamed or temporarily missing;
- search/filter UI hides the fact that an ID is still selected.

### Steps to reproduce
1. Save preferences with a category ID that is not returned in the latest categories fetch.
2. Make another preference change.
3. Save again.

### Expected behavior
Existing selected IDs should be preserved unless the user explicitly removes them.

### Actual behavior
Any selected ID missing from `categories` is omitted from the payload.

### Suggested fix
Send IDs directly if the API supports it, or merge loaded category objects with existing preference objects before saving.

### Acceptance criteria
- Saving does not drop selected preferences that are not in the current category response.
- The behavior is covered by a unit/component test.
- Empty or stale category responses do not wipe valid saved preferences.

Tomeshwari-02:
## Summary
- memoized the saved notification preference topics from the current preferences response
- preserved selected saved topics that are not present in the latest category list when saving

## Why
A partial or stale category response should not cause valid saved preferences to be dropped on the next save.

Fixes #1999

## Validation
- `git diff --check`

### DIFF
--- a/frontend/App.tsx
+++ b/frontend/App.tsx
@@ -7,17 +7,17 @@ import { logger } from './src/services/monitoring/logger';
 import AppContent from './src/components/AppContent';
 
 function App() {
-const [queryClient] = useState(
-() =>
-new QueryClient({
-defaultOptions: {
-queries: {
-retry: 2,
-staleTime: 1000 * 60,
-},
-},
-})
-);
+  const [queryClient] = useState(
+    () =>
+      new QueryClient({
+        defaultOptions: {
+          queries: {
+            retry: 2,
+            staleTime: 1000 * 60,
+          },
+        },
+      }),
+  );
 
   useEffect(() => {
     let isMounted = true;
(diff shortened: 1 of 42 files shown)


SB2318 (changes requested):
This PR currently has too many merge conflicts, so I can't review or merge it in its current state. Please resolve the conflicts with the latest target branch, verify that the project builds successfully, and then request another review. Closing this PR as of now.

Thank for your contribution!

PR #2011, opened 2026-06-28
fix: await user profile refresh

### LINKED ISSUE #2001: [Bug] User profile pull-to-refresh spinner stops before refetch completes
### Problem
`UserProfileScreen` starts the pull-to-refresh spinner, calls `refetch()`, and immediately stops the spinner without waiting for the request to finish. This makes the refresh indicator disappear before new data has loaded.

### Affected code
- `frontend/src/screens/UserProfileScreen.tsx`

Current logic:

```ts
const onRefresh = useCallback(() => {
  setRefreshing(true);
  refetch();
  setRefreshing(false);
}, [refetch]);
```

### Steps to reproduce
1. Open another user's profile on a slow network.
2. Pull to refresh.
3. Observe that the spinner stops almost instantly even though the data request is still running.

### Expected behavior
The refresh indicator should remain visible until `refetch()` resolves or rejects.

### Actual behavior
The indicator stops immediately after calling `refetch()`.

### Suggested fix
Make `onRefresh` async and move `setRefreshing(false)` into a `finally` block:

```ts
const onRefresh = useCallback(async () => {
  setRefreshing(true);
  try {
    await refetch();
  } finally {
    setRefreshing(false);
  }
}, [refetch]);
```

### Acceptance criteria
- Pull-to-refresh spinner stays visible while the request is in flight.
- The spinner stops after success or failure.
- Add a regression test if practical.

Tomeshwari-02:
## Summary
- made user profile pull-to-refresh wait for `refetch()` to finish
- moved spinner reset into a `finally` block so it clears after success or failure

## Why
The refresh indicator should reflect the actual network request instead of stopping immediately after it starts.

Fixes #2001

## Validation
- `git diff --check`

### DIFF
--- a/frontend/App.tsx
+++ b/frontend/App.tsx
@@ -7,17 +7,17 @@ import { logger } from './src/services/monitoring/logger';
 import AppContent from './src/components/AppContent';
 
 function App() {
-const [queryClient] = useState(
-() =>
-new QueryClient({
-defaultOptions: {
-queries: {
-retry: 2,
-staleTime: 1000 * 60,
-},
-},
-})
-);
+  const [queryClient] = useState(
+    () =>
+      new QueryClient({
+        defaultOptions: {
+          queries: {
+            retry: 2,
+            staleTime: 1000 * 60,
+          },
+        },
+      }),
+  );
 
   useEffect(() => {
     let isMounted = true;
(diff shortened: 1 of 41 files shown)


SB2318:
Sorry, @Tomeshwari-02, cannot accept the contribution. First of all, the issue has been closed; secondly, you are not the assignee of the issue.

PR #2012, opened 2026-06-28
fix: use stable trusted user keys

### LINKED ISSUE #2002: [Bug] Trusted users modal uses unstable array index keys for users without _id
### Problem
`TrustedUsersModal` falls back to the array index when a trusted user does not have `_id`. Index keys are unstable when list order changes or items are inserted/removed, which can cause rows to reuse stale UI state.

### Affected code
- `frontend/src/components/TrustedUsersModal.tsx`

Current key extractor:

```tsx
keyExtractor={(item, index) => item._id || String(index)}
```

### Steps to reproduce
1. Open the trusted readers modal with a response where one or more items lack `_id`.
2. Refresh or reorder the trusted users list.
3. React can reuse rows incorrectly because the fallback key is based on position.

### Expected behavior
Every row should use a stable, unique key derived from persistent user data.

### Actual behavior
Rows without `_id` use their array index as the key.

### Suggested fix
Use a stable fallback such as user handle/email/profile ID when available, and avoid rendering rows that cannot be uniquely identified.

### Acceptance criteria
- The modal no longer uses array indexes as persistent list keys.
- Missing identifier cases are handled explicitly.
- Add a small test for fallback key behavior if feasible.

Tomeshwari-02:
## Summary
- replaced array-index fallback keys in the trusted users modal
- added stable fallback key fields and filtered rows that do not have any stable identifier

## Why
FlatList rows should use stable keys so reordered or refreshed trusted-reader data does not reuse stale row state.

Fixes #2002

## Validation
- `git diff --check`

### DIFF
--- a/frontend/App.tsx
+++ b/frontend/App.tsx
@@ -7,17 +7,17 @@ import { logger } from './src/services/monitoring/logger';
 import AppContent from './src/components/AppContent';
 
 function App() {
-const [queryClient] = useState(
-() =>
-new QueryClient({
-defaultOptions: {
-queries: {
-retry: 2,
-staleTime: 1000 * 60,
-},
-},
-})
-);
+  const [queryClient] = useState(
+    () =>
+      new QueryClient({
+        defaultOptions: {
+          queries: {
+            retry: 2,
+            staleTime: 1000 * 60,
+          },
+        },
+      }),
+  );
 
   useEffect(() => {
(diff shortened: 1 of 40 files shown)


SB2318:
Sorry @Tomeshwari-02 Cannot accept the contribution. First of all, the issue has been closed; secondly, you are not the assignee of the issue.

PR #2013, opened 2026-06-28
fix: avoid delayed podcast resume prompts

### LINKED ISSUE #2003: [Bug] Podcast resume prompt can interrupt after user starts playback
### Problem
`PodcastDetail` schedules the resume-position prompt 500ms after mount. If the user starts playback, pauses, or navigates quickly before that timeout runs, the delayed resume alert can still appear and interrupt the action the user just took.

### Affected code
- `frontend/src/screens/PodcastDetail.tsx`

Relevant flow:

```ts
const timeout = setTimeout(() => {
  void checkResume();
}, 500);
```

`checkResume()` shows an `Alert` when saved progress exists.

### Steps to reproduce
1. Open a podcast with saved playback progress.
2. Quickly tap play or another control before the delayed resume prompt appears.
3. The delayed prompt can appear after the user has already made a playback decision.

### Expected behavior
The resume prompt should not interrupt after the user has already interacted with playback controls.

### Actual behavior
The prompt is only delayed/cancelled on unmount, not on user playback interaction.

### Suggested fix
Track whether the user has interacted with playback before the prompt fires, and skip the prompt if they have. Alternatively, show the prompt before enabling playback controls.

### Acceptance criteria
- The resume prompt appears only when the user has not already chosen a playback action.
- The timeout is cleared or ignored once playback controls are used.
- Saved-position resume behavior still works on normal screen entry.

Tomeshwari-02:
## Summary
- tracked user playback interaction on podcast detail controls
- skipped the delayed saved-position prompt when the user has already interacted with playback

## Why
The resume prompt should not interrupt users after they have already chosen to play, pause, or seek.

Fixes #2003

## Validation
- `git diff --check`

### DIFF
--- a/frontend/App.tsx
+++ b/frontend/App.tsx
@@ -7,17 +7,17 @@ import { logger } from './src/services/monitoring/logger';
 import AppContent from './src/components/AppContent';
 
 function App() {
-const [queryClient] = useState(
-() =>
-new QueryClient({
-defaultOptions: {
-queries: {
-retry: 2,
-staleTime: 1000 * 60,
-},
-},
-})
-);
+  const [queryClient] = useState(
+    () =>
+      new QueryClient({
+        defaultOptions: {
+          queries: {
+            retry: 2,
+            staleTime: 1000 * 60,
+          },
+        },
+      }),
+  );
 
   useEffect(() => {
     let isMounted = true;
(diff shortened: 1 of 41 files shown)


SB2318:
Sorry @Tomeshwari-02 Cannot accept the PR; there are so many merge conflicts. Secondly, although the assignee takes a little longer, you are not the assignee of the issue.

PR #2020, opened 2026-06-29
fix: sync article card state

### LINKED ISSUE #2005: [Bug] ArticleCard like/save/repost UI can stay stale after item prop updates
### Problem
`ArticleCard` initializes several local UI states from the incoming `item` prop, but those states are not synchronized when the parent refetches or replaces the article item. This can leave like/save/repost UI stale after data updates.

### Affected code
- `frontend/src/components/ArticleCard.tsx`

Examples:

```ts
const [isLiked, setIsLiked] = useState(/* derived from item.likedUsers */);
const [likeCount, setLikeCount] = useState(item.likedUsers.length);
const [saved, setSaved] = useState(item.savedUsers.includes(user_id));
const [reposted, setReposted] = useState(/* derived from item.repostUsers */);
```

These values are only initialized on first render.

### Steps to reproduce
1. Render an article card.
2. Trigger a refetch or parent update that changes `item.likedUsers`, `item.savedUsers`, or `item.repostUsers`.
3. Observe that the card can continue showing the old local state.

### Expected behavior
The card should reflect the latest article data from props after refetches or parent updates.

### Actual behavior
Local state can remain stale because it is derived from props only once.

### Suggested fix
Add `useEffect` synchronization for prop-derived states, or move these values into a reducer that explicitly reconciles server data with optimistic updates.

### Acceptance criteria
- Like count, liked state, saved state, and repost state update when the `item` prop changes.
- Existing optimistic updates still work.
- Add a component test that rerenders with updated article props.

Tomeshwari-02:
## Summary
- synchronized ArticleCard like, save, repost, and count state when the incoming article item changes

## Why
ArticleCard initializes local UI state from props. When a parent refetches or replaces the item, the UI should reflect the latest server-backed data instead of keeping stale local values.

Fixes #2005

## Validation
- `git diff --check`

### DIFF
--- a/frontend/App.tsx
+++ b/frontend/App.tsx
@@ -7,17 +7,17 @@ import { logger } from './src/services/monitoring/logger';
 import AppContent from './src/components/AppContent';
 
 function App() {
-const [queryClient] = useState(
-() =>
-new QueryClient({
-defaultOptions: {
-queries: {
-retry: 2,
-staleTime: 1000 * 60,
-},
-},
-})
-);
+  const [queryClient] = useState(
+    () =>
+      new QueryClient({
+        defaultOptions: {
+          queries: {
+            retry: 2,
+            staleTime: 1000 * 60,
+          },
+        },
+      }),
+  );
 
   useEffect(() => {
     let isMounted = true;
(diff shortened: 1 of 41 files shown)


SB2318 (commented):
Thanks for your contribution @Tomeshwari-02. But you are not the assignee of this issue.

PR #2031, opened 2026-06-29
fix: propagate React Query errors instead of returning fallback values

### LINKED ISSUE #1147: ## Bug: 10+ React Query hooks catch errors and return `null` — error states never shown to users
### Description
At least 10 React Query `queryFn` implementations wrap `axios.get` in try-catch blocks and return `null` on error instead of letting the error propagate. React Query never sets `isError = true`, so consuming screens show empty/loading states instead of error messages when API calls fail.

### Steps to Reproduce
1. Turn off network connectivity
2. Navigate to any screen that uses one of the affected hooks (Notifications, Podcasts, Articles, etc.)
3. The screen shows an empty state or infinite loading instead of an error message
4. Check the Redux DevTools or React Query DevTools — `isError` is `false` and `data` is `null`

### Affected Hooks

| File | Returns on Error |
|------|-----------------|
| `frontend/src/hooks/useGetAllNotifications.ts` | `null` |
| `frontend/src/hooks/useGetAllPodcasts.ts` | `null` |
| `frontend/src/hooks/useGetArticleTags.ts` | `null` |
| `frontend/src/hooks/useGetDiscardedPodcast.ts` | `null` |
| `frontend/src/hooks/useGetPaginatedArticles.ts` | `null` |
| `frontend/src/hooks/useGetPendingPodcasts.ts` | `null` |
| `frontend/src/hooks/useGetReportReasons.ts` | `null` |
| `frontend/src/hooks/useGetSinglePodcastDetails.ts` | `null` |
| `frontend/src/hooks/useGetUnreadNotificationCount.ts` | `0` |
| `frontend/src/hooks/useGetUserPublishedPodcasts.ts` | `null` |

### Actual Behavior (representative — `useGetPaginatedArticles.ts`)

```typescript
queryFn: async () => {
  try {
    const response = await axios.get(`${PROD_URL}/articles?page=${page}`);
    return response.data as ArticleRes;
  } catch (err) {
    console.error('Error fetching articles:', err);
    return null;   // BUG: React Query sees this as success with data=null
  }
},
```

### Expected Behavior
Errors should propagate to React Query so `isError = true` and the `error` object is populated. Consuming components can then show error states.

### Implementation Hints
Remove the try-catch wrapper and let errors throw naturally:
```typescript
queryFn: async () => {
  const response = await axios.get(`${PROD_URL}/articles?page=${page}`);
  return response.data as ArticleRes;
},
```

If error logging is needed, use React Query's `onError` callback or a global Axios interceptor instead.

### Affected Files
- `frontend/src/hooks/useGetAllNotifications.ts`
- `frontend/src/hooks/useGetAllPodcasts.ts`
- `frontend/src/hooks/useGetArticleTags.ts`
- `frontend/src/hooks/useGetDiscardedPodcast.ts`
- `frontend/src/hooks/useGetPaginatedArticles.ts`
- `frontend/src/hooks/useGetPendingPodcasts.ts`
- `frontend/src/hooks/useGetReportReasons.ts`
- `frontend/src/hooks/useGetSinglePodcastDetails.ts`
- `frontend/src/hooks/useGetUnreadNotificationCount.ts`
- `frontend/src/hooks/useGetUserPublishedPodcasts.ts`

### Labels
`gssoc:approved`, `GSSoC26`, `level:intermediate`, `type:bug`

tylrx404:
## Summary

Fixes #1147

This PR resolves the React Query error propagation issue by allowing React Query to handle request failures instead of returning fallback values.

### Changes

* Removed error-swallowing `try/catch` blocks from the affected React Query `queryFn` implementations.
* Removed fallback `return null` / `return 0` values that caused failed requests to be treated as successful queries.
* Allowed Axios errors to propagate naturally so React Query can correctly populate `isError` and `error`.
* Updated the affected hook return types to match the new behavior.

### Result

Failed API requests are now handled by React Query as intended, allowing consuming components to detect failures through `isError` instead of receiving fallback values.

### Scope

* Updated only the 10 hooks listed in the issue.
* No changes to API endpoints.
* No changes to query keys.
* No changes to business logic or UI behavior.


SB2318 (changes requested):
Hey @tylrx404 Thanks for your contribution! But please mention a valid issue number with your PR for further proceed

tylrx404:
Hi @SB2318 
Thanks for the review. I checked the issue number referenced in my PR (#1147), but it no longer exists/is inaccessible. Could you please let me know the correct issue number that this PR should be linked to? I'll update the PR description immediately.

PR #2033, opened 2026-06-30
fix: implement silent token refresh on 401 instead of force logout

### LINKED ISSUE #1981: [Bug] REFRESH_TOKEN_API is dead code — 401 always force-logs out users instead of silently refreshing
## Problem

`REFRESH_TOKEN_API` (`/user/refreshToken`) is defined in `APIUtils.ts` and the login response already stores a `refreshToken`, but the 401 interceptor **never calls it**. On any 401 response, the app immediately clears the session and drops the user into guest mode with no refresh attempt:

```ts
// frontend/src/helper/setupAxiosInterceptor.ts (lines 62–88)
if (error?.response?.status === 401) {
  store.dispatch(setUserToken(''));
  store.dispatch(setUserId(''));
  store.dispatch(setUserHandle(''));
  store.dispatch(setGuestMode(true));

  secureRemoveItem(SECURE_KEYS.USER_TOKEN);
  removeItem(KEYS.USER_TOKEN_EXPIRY_DATE);
  // shows "Session Expired" toast — no call to REFRESH_TOKEN_API
}
```

Additionally, `LoginScreen.tsx` (line 158) stores `data?.refreshToken` as the bearer sent with every subsequent request — meaning the refresh token *is* the session credential. There is no separate access token in play, which makes silent refresh even more critical: expiry maps directly to a forced logout.

## Impact

- Every token expiry is a forced logout, even when a perfectly valid refresh token is in secure storage.
- `REFRESH_TOKEN_API` (`/user/refreshToken`) is unreachable dead code — the backend endpoint exists, the constant is defined, but no code path ever calls it.
- Users with long sessions lose context without warning and must log in again unnecessarily.
- The `refreshToken` field on the `User` type (`type.ts:737`) and the `AuthData` storage in `LoginScreen` are both wasted — they are stored but never read back for refresh.

## Steps to Reproduce

1. Log in to the app.
2. Wait for the token to expire (or temporarily set a very short expiry on the backend).
3. Perform any authenticated action — load profile, like an article, open a podcast.
4. Observe: user is immediately shown "Your session has expired. You are now browsing as a guest." with no background refresh attempt.

## Expected Behaviour

On a 401 response, the interceptor should:

1. Retrieve the stored refresh token from `SecureStorage`.
2. Call `REFRESH_TOKEN_API` with it.
3. On success → store the new token, retry the original failed request transparently.
4. Only on refresh failure (expired or invalid refresh token) → clear session and show "Session Expired".

This is the standard silent-refresh pattern. The backend endpoint already exists; only the client-side call is missing.

## Proposed Fix

1. Add a module-level `isRefreshing` flag and a `failedQueue` array in `setupAxiosInterceptor.ts` to coalesce concurrent 401s during a refresh (prevents duplicate refresh calls when multiple requests fail simultaneously).
2. In the 401 branch of `handleError`, call `REFRESH_TOKEN_API` with the stored refresh token before clearing the session.
3. On a successful refresh, store the new token via `secureStoreItem` and retry queued requests.
4. Only fall through to the existing logout/guest-mode logic if the refresh call itself returns 401 or fails.

## Files Affected

- `frontend/src/helper/setupAxiosInterceptor.ts` — 401 handler needs the refresh call added
- `frontend/src/helper/APIUtils.ts` — `REFRESH_TOKEN_API` constant already defined (line 33), needs to be exported
- `frontend/src/helper/SecureStorageUtils.ts` — `secureRetrieveItem` already available for reading the stored token

**Complexity**: Level 2 | **Program**: GSSOC '26

trimoyee-g:
## Summary

Fixes #1981

`REFRESH_TOKEN_API` (`/user/refreshToken`) was defined in `APIUtils.ts` and the login response already stored a `refreshToken`, but the 401 interceptor never called it — any expired token immediately force-logged the user out into guest mode.

- Adds a `isRefreshing` flag + `failedQueue` to coalesce concurrent 401s so only one refresh call is made at a time
- On 401: reads the stored token, POSTs to `REFRESH_TOKEN_API`, stores the new token, and retries the original request transparently
- Only falls back to `clearSession()` (the existing "Session Expired" flow) if the refresh itself fails
- Extracts session-clearing logic into `clearSession()` to avoid duplication
- Fixes a potential `TypeError` by guarding `originalRequest &&` before accessing `._retry`

## Files Changed

- `frontend/src/helper/setupAxiosInterceptor.ts`

## Test Plan

- [ ] Log in, wait for token expiry (or shorten expiry on backend), perform an authenticated action — should silently refresh and succeed without showing "Session Expired"
- [ ] With an expired AND invalid refresh token — should show "Session Expired" and drop to guest mode as before
- [ ] Rapid concurrent requests during token expiry — only one refresh call should fire; others should queue and retry with the new token

## Related

Closes #1981

**Program**: GSSOC '26


trimoyee-g:
Hey @SB2318, the initial automated feedback has been addressed. Could you please review the PR? Thanks!

SB2318 (changes requested):
Hello @trimoyee-g Thanks for your contribution! 😇 Our bot has addresses some critical regression regarding access token storage. please carefully read those and please store the long live token.

let me know if you need any help.

PR #2107, opened 2026-07-11
fix: import PermissionsAndroid and fix broken permission check in Utils.ts

### LINKED ISSUE #2099: 🐞[Bug]: PermissionsAndroid used without import in Utils.ts — crashes on Android
### Related Area

Frontend

### Describe the bug

`src/helper/Utils.ts:342-358` references `PermissionsAndroid` but the file only
imports `Alert`, `Linking`, and `Platform` from `react-native`. Since
`PermissionsAndroid` is never imported, calling this function throws a
`ReferenceError` at runtime on Android, crashing the app.

There are two additional bugs in the same block:
1. `PermissionsAndroid.requestMultiple([])` is called with an **empty array**,
   so it requests zero actual permissions — the call is a no-op even once the
   import is fixed.
2. The function ends with a bare `return;`, so it resolves to `undefined`
   instead of a `boolean`. Any caller doing `if (await checkPermission())`
   style logic will get `undefined` instead of a real pass/fail result.

### Expected behavior

- `PermissionsAndroid` should be imported from `react-native`.
- `requestMultiple` should be passed the actual permission constants the
  function is meant to request.
- The function should return a proper `boolean` reflecting whether the
  permission(s) were granted.

### Add ScreenShots

_No response_

_No response_

### Record

anushkasrvstv:
# Pull Request

## Description
Fixes three bugs in `frontend/src/helper/Utils.ts` around Android permission
handling:

1. **Missing import** — `PermissionsAndroid` was used but never imported
   from `react-native`, causing a `ReferenceError` and crash on Android.
2. **No-op permission request** — `requestMultiple([])` was called with an
   empty array, so it requested zero permissions even after the import fix.
   Now requests `WRITE_EXTERNAL_STORAGE` and `READ_EXTERNAL_STORAGE` for
   Android versions below API 33.
3. **Incorrect return type** — the function ended with a bare `return;`,
   resolving to `undefined` instead of a boolean. Callers using
   `if (await checkPermission())`-style logic were getting `undefined`
   rather than a real pass/fail result. Now returns `true` only when both
   permissions are granted.

Closes #2099

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist:
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes


SB2318 (commented):
Thanks for your contribution! But, No need to check for the external storage permission. Please remove it, as it's unnecessary for this implementation.

PR #2214, opened 2026-07-29
fix: add radix parameter to parseInt calls

saurabhhhcodes:


SB2318 (changes requested):
This PR introduces catastrophic regressions and must be reverted. The most critical issue is the wholesale deletion of core application functionality across multiple files, which will render the application completely unusable.

SB2318:
Sorry, @saurabhhhcodes, please research more, fix the regressions, and create a PR. 👍

PR #2266, opened 2026-08-04
fix: debounce refetch & guard in-flight mutations in ArticleScreen actions

### LINKED ISSUE #1955: [Performance] ArticleScreen like/follow/save handlers call refetch() with no debounce — rapid taps fire duplicate sockets and queries
### Description

after reviewing the article interaction handlers, i found that like/follow/save mutations call refetch() immediately on success with no guard against in-flight mutations. mashing the like button triggers overlapping refetches and duplicate notification socket emits.

### The Problem

ArticleScreen.tsx lines 331-346 (like), 372-396 (follow), 409-430 (save): 3 rapid taps → 3 mutations → 3 refetch() → 3 socket emits → duplicate engagement events recorded. no isPending guard on the buttons.

### Proposed Fix

- read isPending from each mutation hook; guard early: if (likeMutationPending) return
- debounce refetch() (500ms)
- emit socket notification only when likeStatus actually changed
- disable action buttons while any mutation is pending

for labels i think gssoc:approved + level:intermediate + type:performance fits here.

vedant7007:
### Problem
In `frontend/src/screens/article/ArticleScreen.tsx`, the like / follow / save / trust handlers each call `refetch()` in `onSuccess`. A burst of rapid taps therefore fired **one article refetch per tap**. The handlers also had no early guard against starting a new mutation while the previous one was still in flight — a window that can produce overlapping requests and duplicate notification `socket.emit` calls before the button's `disabled` state commits.

### Fix
- **Debounced refetch (500ms, cleared on unmount)** used in all four action `onSuccess` callbacks, so a tap burst triggers **at most one** refetch.
- **Early `if (xMutationPending) return;` guard** in each of the four handlers, as defense-in-depth alongside the existing `disabled={xMutationPending}` already on the buttons.
- The article-load `refetch()` (the `useEffect` on `articleId`) is intentionally left immediate — only the action-triggered refetches are coalesced.

### Notes
- `useRef` / `useCallback` / `useEffect` were already imported; no new deps.
- The action buttons already carried `disabled={xMutationPending}`; this adds the handler-level guard the issue asked for so the protection doesn't depend solely on the disabled state having re-rendered.

Closes #1955

### DIFF
--- a/frontend/src/screens/article/ArticleScreen.tsx
+++ b/frontend/src/screens/article/ArticleScreen.tsx
@@ -212,6 +212,24 @@ const ArticleScreen = ({navigation, route}: ArticleScreenProp) => {
     refetch: refetchContent,
   } = useGetArticleContent(resolvedRecordId);
 
+  // Coalesce refetches from rapid like/follow/save/trust taps. Each mutation's
+  // onSuccess previously fired its own refetch(); a burst produced one network
+  // round-trip per tap. Debounce so at most one refetch runs after taps settle.
+  const refetchTimer = useRef<ReturnType<typeof setTimeout> | null>(null);
+  const debouncedRefetch = useCallback(() => {
+    if (refetchTimer.current) clearTimeout(refetchTimer.current);
+    refetchTimer.current = setTimeout(() => {
+      refetchTimer.current = null;
+      refetch();
+    }, 500);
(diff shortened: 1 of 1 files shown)


SB2318 (approved):
@vedant7007  Thanks for your contribution!

Issue #1949, opened 2026-06-26
💡[Feature]: Setup authentication infrastructure for UltimateHealth Web


avrojitduttaj:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Feature Description

UltimateHealth currently lacks a centralized authentication system. This issue aims to establish the authentication infrastructure that future login providers (Email/Password, Google OAuth, etc.) can build upon.

The goal is not to implement every authentication method, but to create a scalable authentication architecture.

### Use Case

1. Create Authentication Context

Create an Auth Context to manage:

Current user
Authentication status
Loading state
Login
Logout
Register (placeholder)

Example:

`src/context/AuthContext.jsx`

2. Create Authentication Hook

Create a custom hook:

src/hooks/useAuth.js

This hook should expose:

user
login()
logout()
register()
loading
isAuthenticated

3. Create Protected Route Component

Create:

`src/components/auth/ProtectedRoute.jsx`

Responsibilities:

Redirect unauthenticated users
Render children for authenticated users
Display loading state while checking authentication
4. Authentication Pages

Create pages:

```
src/pages/Login.jsx
src/pages/Register.jsx
src/pages/ForgotPassword.jsx
```

These pages may contain placeholder forms until backend authentication is integrated.

5. Folder Structure

Suggested structure:

```
src/
│
├── context/
│   └── AuthContext.jsx
│
├── hooks/
│   └── useAuth.js
│
├── components/
│   └── auth/
│       └── ProtectedRoute.jsx
│
├── pages/
│   ├── Login.jsx
│   ├── Register.jsx
│   └── ForgotPassword.jsx
│
├── services/
│   └── authService.js
│
└── utils/
    └── authHelpers.js
```

6. Routing

Add routes for:

```
/login
/register
/forgot-password
```

Protect private routes using ProtectedRoute.

7. Placeholder Authentication Service

Create a service layer that can later be connected to Supabase, Firebase, Clerk, or another backend.

### Priority

High

### Record

- [x] I have read the Contributing Guidelines
- [ ] I'm a GSSOC'24 contributor
- [x] I'm a GSSOC'26 contributor
- [ ] I'm a IEEE IGDTUW contributor
- [x] I want to work on this issue

utsavimohite-arch:
I am GSSoC 2026 contributor please assign this to me 
/assign

Akriti1331:
@SB2318 Please assign this issue to me.

SB2318:
[...]

Issue #1697, opened 2026-06-17
💡[Feature]: Hardcoded User-Facing Strings Prevent Scalable Localization Across the Application


Vishxlll20:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Feature Description

The application currently contains a large number of hardcoded user-facing strings across multiple screens, including placeholders, validation messages, alerts and error notifications.

Examples found in the codebase:

"Enter your email"
"Password"
"Article not found"
"Failed to upload your post"
"Please select a target language for the translation"
"Search contributors"

These strings are directly embedded in components, making it difficult to implement application-wide localization and maintain consistency across the platform.

### Use Case

- Prepare the application for future multilingual support.
- Improve maintainability by centralizing user-facing text.
- Allow developers to manage translations from a single source.
- Ensure consistent wording across screens and features.

### Benefits

- Easier integration of internationalization (i18n) frameworks.
- Better user experience for multilingual users.
- Reduced duplication of text across components.
- Simplified maintenance and future feature development.
- Supports scalability as more languages are added.

### Add ScreenShots

<img width="1455" height="1079" alt="Image" src="https://github.com/user-attachments/assets/6e83847c-f56b-46f4-9c1d-dfb6cb4616f5" />
<img width="1455" height="1079" alt="Image" src="https://github.com/user-attachments/assets/1da3aa3a-57fb-418a-b4dc-a887ba44e9af" />

<img width="1455" height="1079" alt="Image" src="https://github.com/user-attachments/assets/854c1a2a-1f34-4a8e-bc61-090b55443370" />

### Priority

Medium

### Record

- [x] I have read the Contributing Guidelines
- [ ] I'm a GSSOC'24 contributor
- [x] I'm a GSSOC'26 contributor
- [ ] I'm a IEEE IGDTUW contributor
- [x] I want to work on this issue

madsysharma:
/assign
I would like to work on this as part of GSSoC 2026.

aeriellaaa:
I'd like to work on this, please assign me
as a part of GSSoC2026

Stewartsson:
@SB2318 i am a GSSoC contributor can you please assign this to me

PR #2278, opened 2026-08-05
feat(web): add authentication infrastructure (#1949)

Anijesh:
Closes #1949

## Overview

Establishes the authentication architecture future login providers can build on. Nothing here is a throwaway placeholder — the service layer is wired to the real `/user/*` endpoints from the spec @SB2318 posted in the issue.

**One adaptation worth flagging:** the issue suggests a Vite/React-Router layout (`src/pages/Login.jsx`, `AuthContext.jsx`, router-level `ProtectedRoute`). This app is Next.js App Router + TypeScript, so the same architecture is expressed in its conventions — routes live in `src/app/[locale]/`, the provider is mounted in the locale layout, and `ProtectedRoute` gates a client subtree rather than a router element. Every piece the issue asks for is present, under a Next.js-shaped path.

## Architecture

Each layer is replaceable without touching the one above it.

| Layer | File | Responsibility |
|---|---|---|
| Helpers | `src/lib/auth/auth-helpers.js` | Pure rules mirroring the backend — password policy, handle/name format, OTP shape, response-envelope and token readers |
| Service | `src/lib/auth/auth-service.ts` | The only module that talks to `/user/*` |
| Storage | `src/lib/auth/auth-storage.ts` | Token persistence behind one seam |
| Context | `src/contexts/AuthContext.tsx` | Session source of truth |
| Hook | `src/hooks/useAuth.ts` | Consumer API |
| Guard | `src/components/auth/ProtectedRoute.tsx` | Route gating |
| Types | `src/lib/auth/types.ts` | Shared shapes + `AuthError` |

### Details worth reviewing

**Token field is not stable.** Per the mobile client (`useUserLogin.ts`), the token sits at the top level of the response — never inside `user` — and arrives as `token`, `accessToken` *or* `refreshToken`, sometimes wrapped in an extra `data` envelope. `readAuthSession()` handles all of it in one tested place instead of scattering `?.data?.token ?? …` through the UI.

**Doctor fields are conditional.** `/user/register` requires `qualification`, `specialization` and `Years_of_experience` only when `isDoctor` is true, and `contact_detail` only for doctors. The service omits those keys entirely for readers rather than sending empty strings, and the form reveals the fieldset when the box is ticked.

**Client-side validation mirrors the backend.** The password rule (8+ chars, upper, lower, number, special), the 3–20 char alphanumeric handle and the 2–50 char name are all checked before submitting, so users see the problem without a round trip.

**Sign-out is local-first.** `logout()` clears the stored session in a `finally` block, so a failing `/user/logout` still signs the user out of this browser.

**No stuck loading state.** If the session check fails for any reason the context resolves to `unauthenticated` rather than leaving the app spinning forever.

**Open-redirect guard.** `/login?next=…` is only honoured when it is a same-origin path — `//evil.example.com` and absolute URLs are ignored and the default destination is used.

## Routes added

- **`/register`** — full sign-up with the conditional doctor fieldset.
- **`/forgot-password`** — two-step reset, because the backend splits it across `/user/forgotpassword` (sends a 6-digit OTP) and `/user/verifypassword` (consumes it). The email carries over between steps.
- **`/login`** — refactored to go through the context instead of its own `fetch`; keeps its existing translations and its redirect to `/delete-account`, and now links to the two new pages.

### Using the guard

```tsx
"use client";
import ProtectedRoute from "@/components/auth/ProtectedRoute";

export default function Page() {
  return (
    <ProtectedRoute>
      <Dashboard />
    </ProtectedRoute>
  );
}
```

I did not retrofit it onto `/delete-account`: that page already runs its own sign-in check via `AccountDeletionPage`, and rewiring it is a behaviour change beyond this issue. It is the obvious first adopter once someone touches that flow.

## Screenshots

**Register — with the conditional doctor fields revealed**

Registration form showing name, handle, email, password, contact number, a medical professional checkbox and the revealed professional details fieldset

**Forgot password — step two after the code is sent**

Password reset form asking for the six-digit verification code and a new password

**Login — now linking to register and reset**

Login form with Forgot password and Create an account links below the submit button

## Verification

```
npm test          # 30/30 pass (10 new)
npx tsc --noEmit  # clean
npm run lint      # no new problems; the 2 errors are pre-existing on web
npm run build     # compiles; /register and /forgot-password routed
```

Beyond that, I drove the real pages in Chrome against a stubbed backend — **21/21 checks pass**:

*Register* — empty submit raises 4 field errors and makes no API call · weak password rejected client-side · doctor checkbox reveals the fieldset · valid submit posts to `/user/register` · leading `@` stripped from the handle · doctor-only keys omitted for a reader · returned token persisted

*ProtectedRoute* — signed-out visitor redirected to `/login?next=…` · children never rendered while signed out · signed-in visitor sees the children

*Forgot password* — invalid email blocked client-side · advances to the OTP step · server message surfaced

*Login* — redirects to `?next` · stores the token · **refuses an off-site `?next`** · surfaces the server's error message

The verification used a temporary route to mount `ProtectedRoute`; it was removed before committing and is not in this diff.

## Notes for reviewers

- **New pages use plain English rather than `next-intl` keys.** Only `/login` is translated today; `/contribute`, `/our-contributors` and `/medical-glossary` are all hardcoded English. Adding `Register`/`ForgotPassword` namespaces would mean inventing de/es/fr/hr translations I cannot check, so I matched the majority of the codebase and left localisation to follow #1697. Happy to add the keys if you'd prefer.
- **Token storage is localStorage today.** The backend also sets a session cookie and every request sends `credentials: "include"`, so the token is a convenience for the `Authorization` header rather than the only thing holding the session. It is isolated in `auth-storage.ts` precisely so it can move to an httpOnly-cookie or BFF approach without touching the context or the pages — worth deciding before real private routes ship.
- `npm run knip` still fails on `web` with `Invalid input (unrecognized_keys: $comment)` — pre-existing, and the workflow is `continue-on-error`.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -6,7 +6,7 @@
     "dev": "node --no-warnings node_modules/next/dist/bin/next dev",
     "build": "next build",
     "start": "next start",
-    "test": "node --no-warnings --test src/app/[locale]/admin-agreement/signatureValidation.test.mjs src/app/api/csp-report/sanitizeCspReport.test.mjs src/components/articles/article-layout.test.mjs src/lib/api.test.mjs src/lib/nav-active.test.mjs",
+    "test": "node --no-warnings --test src/app/[locale]/admin-agreement/signatureValidation.test.mjs src/app/api/csp-report/sanitizeCspReport.test.mjs src/components/articles/article-layout.test.mjs src/lib/api.test.mjs src/lib/nav-active.test.mjs src/lib/auth/auth-helpers.test.mjs",
     "lint": "eslint",
     "knip": "knip",
     "knip:deps": "knip --dependencies",
--- a/web/src/app/[locale]/forgot-password/page.tsx
+++ b/web/src/app/[locale]/forgot-password/page.tsx
@@ -4,17 +4,15 @@ import Link from "next/link";
 import { useRouter } from "next/navigation";
 import { useState, type FormEvent } from "react";
 
-import { PageWrapper, Section } from "@/components/layout";
+import AuthShell, {
+  authErrorClass,
+  authFieldClass,
+  authLabelClass,
+  authLinkClass,
+  authSubmitClass,
+} from "@/components/auth/AuthShell";
 import { Alert, AlertDescription, AlertTitle } from "@/components/ui/alert";
 import { Button } from "@/components/ui/button";
-import {
-  Card,
-  CardContent,
-  CardDescription,
-  CardFooter,
-  CardHeader,
-  CardTitle,
-} from "@/components/ui/card";
 import { Input } from "@/components/ui/input";
 import { Label } from "@/components/ui/label";
 import { Spinner } from "@/components/ui/spinner";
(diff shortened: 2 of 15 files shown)


Anijesh:
Thanks for the review — going through each point.

### Addressed

**Unused `isUserHandleAvailable`** — removed. Nothing consumed it, and the "handle already taken" case already surfaces through the normal `/user/register` error path, so a second endpoint call would only duplicate that. Easy to re-add behind a debounced field if we want the check to happen before submit.

### Answered, not changed

**Token in `localStorage` / XSS** — agreed on the risk, and I'd rather not decide this unilaterally. The backend already sets a session cookie and every request here sends `credentials: "include"`, so the stored token is only there to populate an `Authorization` header. That means there are two possible hardenings:

1. keep it in memory only, so a refresh relies purely on the cookie session, or
2. move to an httpOnly cookie behind a BFF route.

Both depend on whether the cookie session alone is sufficient for `/user/getprofile` on web — I verified session restore against a stubbed backend, so I can't confirm that from here, and guessing wrong would silently sign users out on every refresh. `auth-storage.ts` exists as a single seam precisely so this is a one-file change. @SB2318, happy to do either as a follow-up once you confirm which the backend supports.

**No automated tests for the service / context / hook** — a fair gap, but the web app has no React test harness today: `npm test` is `node --test` over `.mjs` files, with no jest/vitest or Testing Library. Adding one is its own infrastructure change and felt out of scope here. What I did instead was drive the real pages in Chrome against a stubbed backend, covering exactly those layers — 21 checks, all passing (redirect-with-`next`, children never rendered while signed out, token persistence, off-site `next` rejected, server errors surfaced, doctor-only fields omitted). Details are in the PR description. If you'd like, I can open a separate issue for adding Vitest + Testing Library to the web app.

**Convert `auth-helpers.js` to TypeScript** — this one is constrained by the test runner rather than preference. `npm test` runs plain `node --test` with no TypeScript transform, so a `.ts` module cannot be imported by the `.mjs` tests. The two existing precedents do the same thing for the same reason: `src/lib/api-url.mjs` (tested by `api.test.mjs`) and `src/components/articles/article-layout.js` (tested by `article-layout.test.mjs`). The file is fully JSDoc-typed, so consumers still get inference and `tsc --noEmit` checks the call sites. Worth revisiting together with the test-harness point above.

**`DEFAULT_REDIRECT` pointing at `/delete-account`** — sensible in general, but this specific page exists to gate account deletion, and its translated copy says so in all five locales: `heading` is "Login Required", `description` is "Please log in to continue to account deletion", `successDescription` is "Redirecting to account deletion". Redirecting to `/` would contradict the text the user just read and would need those strings rewritten in en/de/es/fr/hr. Since any other caller can pass `?next=`, I kept the existing destination as the default. Glad to change it if the page is meant to become a general sign-in entry point.

Anijesh:
Pushed a security fix to this PR — flagging it explicitly since it changes code you may have already looked at.

**The open-redirect guard I described in the PR body was not actually sufficient.** I re-tested it in Chrome against the running app rather than trusting the code, and three payloads got through:

```
/login?next=/\evil.example.com
/login?next=/\/evil.example.com
/login?next=/<TAB>/evil.example.com
```

Each one reached `http://evil.example.com/` after a successful sign-in. The check was `startsWith("/") && !startsWith("//")`, which misses two things: browsers treat a backslash as a path separator in http(s) URLs, and they strip control characters while parsing — so `/\evil.com` and `/<TAB>/evil.com` both become `//evil.com` by the time navigation happens.

Replaced it with `safeRedirectPath()`, which resolves the candidate against the current origin and requires the origins to match, then rejects a resolved path that is itself protocol-relative (`/..//evil.com` resolves same-origin but yields `//evil.com`).

Re-ran the same browser probe against the fix: all eight vectors are now contained and fall back to the default destination, while legitimate targets like `/en/articles?page=2#top` are preserved intact. The browser-confirmed payloads are now a shared vector list with four tests over them, so a refactor can't quietly reopen this. Test count is 34, `tsc` is clean, and the build passes.

Sorry for the churn — better caught here than after merge.

Anijesh:
One more fix, and this one was a real bug rather than hardening.

`register()` was treating the `/user/register` response as a sign-in. It is not — that endpoint returns a short-lived **email-verification** token. The mobile client in this repo settles it:

- `frontend/src/hooks/auth/useUserRegistration.ts` returns `res.data.token`
- `SignUpScreenSecond.tsx` keeps it in local state, posts it to `/user/verifyEmail` as `{email, token}` to send the mail, then navigates to the login screen
- `SECURE_KEYS.USER_TOKEN` is written only by `LoginScreen` — never by a sign-up screen

My context stored that token under `uh.auth.token` and spent it on `/user/getprofile`. Two consequences: every later authenticated request carried a bearer token that isn't a credential, and because nothing in `web/src` called `/user/verifyEmail`, **no verification email was ever sent** — sign-up could not be completed at all. Depending on whether the backend accepted that JWT, the user was either signed in with verification skipped, or left holding a junk token.

`register()` now returns only the verification token, the context posts it to `/user/verifyEmail` and always resolves to `unauthenticated`, and the page reports the outcome instead of redirecting. A failed send no longer shows "registration failed" for an account that was actually created.

@SB2318 — worth confirming the `/user/verifyEmail` contract is POST `{email, token}` as the mobile client uses it; the endpoint list on the issue shows a GET, which I take to be the link inside the email rather than the call that sends it.

37 tests pass, `tsc` clean, lint unchanged, build passes.

SB2318:
> One more fix, and this one was a real bug rather than hardening.
> 
> `register()` was treating the `/user/register` response as a sign-in. It is not — that endpoint returns a short-lived **email-verification** token. The mobile client in this repo settles it:
> 
> * `frontend/src/hooks/auth/useUserRegistration.ts` returns `res.data.token`
> * `SignUpScreenSecond.tsx` keeps it in local state, posts it to `/user/verifyEmail` as `{email, token}` to send the mail, then navigates to the login screen
> * `SECURE_KEYS.USER_TOKEN` is written only by `LoginScreen` — never by a sign-up screen
> 
> My context stored that token under `uh.auth.token` and spent it on `/user/getprofile`. Two consequences: every later authenticated request carried a bearer token that isn't a credential, and because nothing in `web/src` called `/user/verifyEmail`, **no verification email was ever sent** — sign-up could not be completed at all. Depending on whether the backend accepted that JWT, the user was either signed in with verification skipped, or left holding a junk token.
> 
> `register()` now returns only the verification token, the context posts it to `/user/verifyEmail` and always resolves to `unauthenticated`, and the page reports the outcome instead of redirecting. A failed send no longer shows "registration failed" for an account that was actually created.
> 
> @SB2318 — worth confirming the `/user/verifyEmail` contract is POST `{email, token}` as the mobile client uses it; the endpoint list on the issue shows a GET, which I take to be the link inside the email rather than the call that sends it.
> 
> 37 tests pass, `tsc` clean, lint unchanged, build passes.

Give me 30 mins. I will tell u.

SB2318:
@Anijesh, please give some CSS; it looks like my LinkedIn profile. Happy to merge it then with exceptional quality.

Anijesh:
@SB2318 Fair hit — they were unstyled shadcn defaults on a flat grey page. Restyled all three auth routes to the landing page's identity.

**Sign in — light / dark**

Sign-in page in light mode Sign-in page in dark mode

**Sign up — with the doctor fieldset collapsed**

Registration page

**Password reset — dark**

Password reset page in dark mode

What changed:

- A shared `AuthShell` carries the indigo → purple → rose brand identity: soft out-of-focus orbs and a faint grid behind a glassmorphic card, gradient logo mark and CTA, gradient hairline under the heading. One component, so the three routes cannot drift apart.
- Sign-up puts name and handle side by side from `sm` up, and the doctor-only fieldset is a tinted panel that animates in.
- Reset gained a two-step progress indicator and a letter-spaced OTP field.
- Sign-in moved "Forgot password?" up beside the password label.
- Full dark mode, and every entrance animation sits behind `motion-safe:`.

**One thing worth knowing for future web PRs.** `globals.css:87` has an unlayered `* { margin: 0; padding: 0 }` reset, and unlayered CSS outranks Tailwind's layered utilities — so `px-6`, `space-y-5` and friends were computing to **0** here. Same for `h1 { font-size: clamp(2rem,4vw,3rem) }` at `:1985` overriding `text-2xl`. I measured this in the browser rather than guessing, and used Tailwind's `!` modifier as the narrowest fix that keeps everything in utilities. Dropping that global reset would make Tailwind spacing work normally app-wide, but that is a repo-wide change well outside this PR — happy to raise it as its own issue if useful.

Verified after the restyle: validation still renders 4 announced `role="alert"` errors on an empty submit, `?next` still redirects correctly, 37 tests pass, `tsc` clean, lint unchanged at the 2 pre-existing errors, build passes.

SB2318 (approved):

Anijesh:
image
@SB2318  plz add (gssoc:approved) label otherwise it won't count

SB2318:
@Anijesh please check

Anijesh:
image


@SB2318 bro check this, still same

SB2318:
> image
> @SB2318 bro check this, still same

cache data

SB2318:
@Anijesh Here are your 6 closed PRs

image

Issue #1174, opened 2026-06-06
[Performance/Next.js] Critical LCP Degradation & Hydration Mismatch in Web App


sonusharma6-dsa:
### Current Behavior

The web frontend (Next.js) suffers from a massive main-thread blockage during the React 18 hydration phase. The server injects a massive inline JSON payload (__NEXT_DATA__) containing the entire initial Redux state. On mobile devices, parsing this JSON and executing the hydration blocks the main thread, delaying the Time to Interactive (TTI) by over 2.5 seconds and severely degrading the Largest Contentful Paint (LCP) metric.

### Proposed Solution

1. **State Slicing:** Only serialize the critical path data needed for the initial viewport. Defer non-critical state fetching to client-side useEffect hooks.
2. **Selective Hydration:** Wrap below-the-fold components (like the footer, related articles, and comments section) in <Suspense> boundaries to prioritize hydrating the main article content first.
3. **Streaming SSR:** Leverage Next.js app router streaming features to progressively send HTML chunks to the browser.

sonusharma6-dsa:
/assign
I want to work on this issue, Thanks!

SB2318:
Thank you for the suggestion.

This report does not provide sufficient evidence to demonstrate that the described performance issue exists in the current codebase.

The issue attributes degraded TTI and LCP to hydration of a large `__NEXT_DATA__` payload, but does not include:

* Lighthouse or Web Vitals measurements.
* Chrome Performance recordings.
* Bundle analysis results.
* Serialized state size measurements.
* Affected pages or routes.
* Code locations responsible for the alleged bottleneck.

Additionally, the proposed solutions introduce significant architectural changes, including state restructuring, hydration strategy changes, and streaming SSR adoption. Such changes require evidence that the current implementation is the source of the problem and that the proposed approach would provide measurable benefits.

Please provide reproducible profiling data, affected pages, relevant code locations, and performance measurements demonstrating the issue within this repository.

Until then, this remains an unverified optimization suggestion rather than an actionable feature request or confirmed performance issue.

Stewartsson:
@SB2318  hi I am a GSSoC contributor and I would like to work on this issue please can you assign this to me

Stewartsson:
@SB2318 hi I am a GSSoC contributor and I would like to work on this issue please can you assign this to me

PR #2299, opened 2026-08-07
perf(web): cut render-blocking CSS and hydration work on the landing page

Anijesh:
Fixes #1174

## Context

The issue as filed attributed the problem to a large `__NEXT_DATA__` Redux payload. That diagnosis does not apply to this codebase — there is no Redux here, and the App Router does not emit `__NEXT_DATA__`. @SB2318 was right to ask for evidence before accepting the proposed rewrite.

So I profiled the page instead of assuming, found four concrete causes, and fixed those. No streaming-SSR adoption, no state restructuring — the two proposals from the issue that would have been architectural changes on an unverified premise.

## How this was measured

Production build (`next build` / `next start`), `/en` landing page, headless Chrome over CDP with **4× CPU throttling and ~1.6 Mbps / 150 ms RTT** (the mobile profile the issue describes). Both builds were served **simultaneously on two ports and sampled in interleaved rounds**, so machine noise hits both arms equally. 7 rounds after a discarded warm-up, median reported.

| Metric | Before | After | Delta |
| --- | --- | --- | --- |
| **Largest Contentful Paint** | 3700 ms | 3464 ms | **−236 ms (−6.4%)** |
| **Main-thread blocking (long tasks)** | 561 ms | 450 ms | **−111 ms (−19.8%)** |
| First Contentful Paint | 1708 ms | 1756 ms | +48 ms (within noise) |
| Render-blocking CSS | 481.7 KB (88.4 KB gz) | 325.7 KB (63.4 KB gz) | **−156.0 KB (−25.0 KB gz)** |
| First-load JS | 871.0 KB (263.7 KB gz) | 819.7 KB (248.9 KB gz) | **−51.3 KB (−14.8 KB gz)** |
| Total bytes | 1469.5 KB (366.9 KB gz) | 1310.7 KB (334.3 KB gz) | **−158.8 KB (−32.6 KB gz)** |

The LCP element is the same in both (`p.uh-hero-subtitle`). Run-to-run spread also tightens: LCP ranged 3680–4376 ms before and 3444–3656 ms after.

Two numbers I want to be upfront about:

- **FCP is ~48 ms slower**, which is inside the noise band of this harness (individual FCP samples ranged 1672–2720 ms before and 1704–2608 ms after). I did not want to claim it as an improvement or hide it.
- **The HTML document grows 116.7 KB → 165.2 KB** (14.8 → 22.0 KB gzipped). This is the expected trade: markup that used to be generated by client JavaScript during hydration is now serialized in the RSC payload instead. It is a net win because those bytes no longer have to be parsed, compiled and executed as JavaScript, which is what the long-task reduction reflects.

## What was actually wrong

**1. Tailwind was compiled and shipped twice.** `globals.css` and `globals2.css` each began with `@import "tailwindcss"`, and `layout.tsx` imported both — so every page loaded two near-identical ~200 KB stylesheets, both render-blocking. `globals2.css` is now imported from `globals.css` and the framework is emitted once. This is the single biggest item: −156 KB of render-blocking CSS on every route in the app, not just the landing page.

**2. Inter was declared twice via `next/font`.** Two `Inter()` calls (`--font-inter` and `--font-heading`) produced two identical `@font-face` sets and duplicated font files. `--font-heading` is referenced nowhere in any stylesheet, so one instance now backs the remaining variable.

**3. The entire 1340-line landing page was one `"use client"` component.** Every static section — hero, feature grids, moderator cards, programs, footer — shipped as JavaScript and had to hydrate before the page became interactive. `page.tsx` is now a server component composing server-rendered sections, with client islands only where there is real interactivity:

| Client island | Why it needs to be one |
| --- | --- |
| `SiteHeader` | scroll state, scroll-spy, mobile menu |
| `ScreenshotGallery` | carousels + lightbox |
| `ContactForm` / `NewsletterForm` | form state |
| `TestFlightCta` | owns its own dialog |
| `ScrollReveal` / `DnaCursor` | render `null`, effects only |

`HeroAndDownload` is now server-rendered, so the LCP content no longer waits on the page bundle. It previously took two callbacks to open modals; `onShowComingSoon` was threaded into `AndroidStoreButtons`, **which never used it** — that modal was unreachable, so it is removed. `onJoinTestFlight` became `TestFlightCta`, which owns its own dialog.

**4. The features grid had a hardcoded 1.5 s fake loading state.** `setTimeout(() => setFeaturesLoading(false), 1500)` rendered six skeleton placeholders, then swapped in nine static, already-known cards — a guaranteed layout shift and 1.5 s of empty space for content that needs no fetching. The cards now render in the server HTML.

**Also:** the DNA cursor trail ran a `requestAnimationFrame` loop for the entire session *even when switched off*, clearing a full-viewport canvas every frame. (The toggle that enables it isn't currently rendered anywhere, so this was pure overhead for every visitor.) The loop now only runs while the effect is enabled; the stored preference and its custom event are unchanged.

## Verifying nothing visually changed

I diffed the server-rendered markup of both builds — tag sequence, visible text, and class lists. **The only difference in the entire page** is the features grid rendering its 9 real cards instead of 6 pulsing skeletons:

```
123a124,141
> AI Health Chat Assistant
> Instant, AI-powered health guidance available 24/7.
> Centralized Library
...
```

Everything else is positionally identical. The roadmap card hover was `onMouseEnter`/`onMouseLeave` writing inline styles; it is now a CSS `:hover` rule driven by a per-card `--uh-upcoming-glow` custom property, which renders the same — the inline `transform` was already being overridden by the `uh-fade-up` animation's `both` fill mode, so only the box-shadow was ever visible, and that is preserved.

## Checklist

- [x] `npx tsc --noEmit` — clean
- [x] `npm run lint` — 13 problems (3 errors, 10 warnings) before → **10 problems (3 errors, 7 warnings)** after. All 3 errors are pre-existing and untouched (`ui/navbar.tsx`, `i18n/request.ts`, `ui/cookie-consent.tsx`)
- [x] `npm run build` — succeeds
- [x] No console errors or warnings during page load (checked over CDP across all runs, both builds)
- [x] Dark mode unaffected — both stylesheets already resolved `dark:` to `.dark *` (via `darkMode: "class"` and `@custom-variant`), so merging them changes nothing
- [x] Single `<h1>` per page and semantic tags preserved — markup is unchanged
- [x] No new `any`; components typed
- [x] Single commit, no unrelated files

## Reviewing this

The diff looks large because a 1340-line file was split, but almost all of it is **moved verbatim**. The genuinely new logic is small: `TestFlightCta.tsx`, `ScrollReveal.tsx`, the `DnaCursor.tsx` loop guard, and the `.uh-upcoming-card:hover` rule. The markup-equivalence diff above is the fastest way to confirm the move was faithful.

Happy to split this into separate PRs (CSS dedupe / page split / skeleton removal) if you'd prefer to land them independently — the CSS fix alone is 2 lines and accounts for most of the byte reduction.


Anijesh:
### Functional verification

Since this PR moves a lot of code across the server/client boundary, byte counts alone aren't enough — I drove both builds through an interaction script over CDP to prove the client islands still behave identically. Both servers were running simultaneously and got the same script.

**Refactored build: 25/25 passed, zero console errors or warnings** (i.e. no hydration mismatches).

```
✓ hero h1 rendered                    ✓ lightbox opens
✓ 9 feature cards in DOM immediately  ✓ body scroll locked
✓ no skeleton placeholders            ✓ arrow navigates lightbox
✓ footer rendered                     ✓ Esc closes lightbox
✓ programs rendered                   ✓ body scroll restored
✓ moderator cards rendered            ✓ admin carousel expands
✓ roadmap cards rendered              ✓ TestFlight dialog opens
✓ mobile menu closed initially        ✓ TestFlight dialog labelled
✓ mobile menu opens on click          ✓ Esc closes TestFlight dialog
✓ mobile menu closes again            ✓ contact form validates on input
✓ header gains .scrolled              ✓ newsletter shows empty-state error
✓ gallery slides rendered (30)        ✓ scroll reveal marks elements visible
✓ no DNA canvas when disabled
```

**The same script against the current `web` branch: 23/25.** The only two it fails are the two things this PR deliberately changes:

- `roadmap cards rendered` — the `.uh-upcoming-card` class is new; on `web` the roadmap cards have no class, their hover was inline styles written by JS handlers.
- `no DNA canvas when disabled` — on `web` the DNA canvas **is** in the DOM with its `requestAnimationFrame` loop running, even though the effect is switched off and the toggle that would enable it isn't rendered anywhere. That's the wasted per-frame work this PR removes, now confirmed rather than just claimed.

Every other behaviour — menus, scroll state, carousels, lightbox navigation, both forms' validation, dialog focus/escape handling, scroll-reveal — passes identically on both.

---

Re: the automated reviewer's suggestion of Lighthouse CI / Playwright performance tests in CI — I agree that's the right way to stop these regressions coming back, and it's how the duplicate Tailwind import would have been caught. I've deliberately kept it out of this PR because adding a CI performance harness is an infrastructure change with its own trade-offs (runner cost, threshold tuning, flake budget) that deserves its own review rather than riding along with a page refactor. Happy to open a follow-up issue proposing it, or to add it here if you'd rather it land together — your call, @SB2318.

Also — could someone with triage rights add the **`gssoc`** label? This is GSSoC work against issue #1174 (which carries the label), but I don't have permission to set labels myself.

SB2318 (approved):

PR #2326, opened 2026-08-09
feat: add offline reading mode for recently viewed articles (#1880)

### LINKED ISSUE #1880: 💡 [FEATURE] Add offline reading mode that caches recently viewed articles for reading without internet

## Is there an existing issue for this?

## Describe the feature
Many users in regions with poor internet connectivity 
cannot reliably read health articles when offline. 
Adding an offline reading mode that automatically 
caches the last 10-20 viewed articles using 
AsyncStorage would allow users to access important 
health information even without an internet connection.

## Why is this needed?
UltimateHealth aims to deliver trusted health content 
worldwide including regions with limited connectivity. 
Offline access is critical for a health platform to 
truly serve global users. This aligns directly with 
the project's mission of multilingual worldwide content.

## Proposed Solution
Automatically cache article content to AsyncStorage 
when articles are viewed. Show an "Available Offline" 
badge on cached articles. Display cached articles 
from storage when the device is offline with a 
banner indicating offline mode.

## Acceptance Criteria
- [ ] Last 20 viewed articles cached automatically
- [ ] Offline articles accessible without internet
- [ ] "Available Offline" badge on cached articles
- [ ] Offline mode banner shown when no connection
- [ ] Cache clears articles older than 7 days
- [ ] Works on both Android and iOS

## Is there an existing issue for this?

## Describe the feature
Many users in regions with poor internet connectivity 
cannot reliably read health articles when offline. 
Adding an offline reading mode that automatically 
caches the last 10-20 viewed articles using 
AsyncStorage would allow users to access important 
health information even without an internet connection.

## Why is this needed?
UltimateHealth aims to deliver trusted health content 
worldwide including regions with limited connectivity. 
Offline access is critical for a health platform to 
truly serve global users. This aligns directly with 
the project's mission of multilingual worldwide content.

## Proposed Solution
Automatically cache article content to AsyncStorage 
when articles are viewed. Show an "Available Offline" 
badge on cached articles. Display cached articles 
from storage when the device is offline with a 
banner indicating offline mode.

## Acceptance Criteria
- [ ] Last 20 viewed articles cached automatically
- [ ] Offline articles accessible without internet
- [ ] "Available Offline" badge on cached articles
- [ ] Offline mode banner shown when no connection
- [ ] Cache clears articles older than 7 days
- [ ] Works on both Android and iOS

### Benefits

_No response_

### Add ScreenShots

_No response_

### Priority

High

### Record

Anijesh:
Fixes #1880

## Problem

`useGetArticleDetails` and `useGetArticleContent` fetch over the network on every open, and nothing is retained locally. A user who opened an article minutes ago cannot re-read it once connectivity drops — the query simply errors.

## Acceptance criteria

| Criterion | Status |
|---|---|
| Last 20 viewed articles cached automatically | ✅ `ArticleCacheUtils`, LRU on view |
| Offline articles accessible without internet | ✅ fallback in both read hooks |
| "Available Offline" badge on cached articles | ✅ `ArticleCard` meta row |
| Offline mode banner shown when no connection | ✅ **already existed** — see below |
| Cache clears articles older than 7 days | ✅ TTL pruned on read and write |
| Works on both Android and iOS | ✅ AsyncStorage, no platform-specific code |

The banner criterion needed no work: `NetworkBanner` already renders "No Internet Connection", is mounted in `AppContent.tsx:188`, and has its own test suite. I left it alone rather than duplicating it.

## Changes

**`lib/storage/ArticleCacheUtils.ts`** — AsyncStorage-backed cache, bounded two ways: 20 most-recent entries and a 7-day TTL. Re-viewing an article moves it to the front, so actively read articles survive eviction while forgotten ones age out. Corrupt storage resets itself instead of throwing, so one bad write cannot permanently break offline reading.

**`useGetArticleDetails` / `useGetArticleContent`** — cache on success, serve the cached copy when the request fails offline. The body is fetched separately by PocketBase record id, so it is matched back to the article carrying the same `pb_recordId`; this means the full article is readable offline, not just its metadata. A body with no cached article is not stored, since it would be unreadable anyway.

**`lib/api/networkError.ts`** — `isOfflineError` gates the fallback to connectivity failures only. This is a deliberate call: a 4xx is an answer the server actually gave, and quietly serving stale cached content in its place would hide real errors, including an article taken down upstream. For health content that seemed worth being strict about, so a 404 still surfaces as an error even when a cached copy exists.

**`ArticleCard`** — "Available Offline" badge, following the existing conditional pattern used by the "Trusted by" indicator, with an `accessibilityLabel`.

## Testing

15 new tests across two files, all passing:

- `ArticleCacheUtils.test.ts` (10) — eviction past 20, LRU refresh surviving eviction, no duplicate entries, 7-day expiry, body matching by record id, body preserved across re-view, orphan body ignored, corrupt-data recovery, clear.
- `useOfflineArticleReading.test.ts` (5) — caches automatically on view, serves article and body offline, still fails offline when never viewed, and **does not mask a 404 with stale cached content**.

Full suite: **141 suites, 524 tests, all passing**, `jest` exit code 0. `tsc --noEmit` clean for the touched files, `eslint` clean.

## One known nit

With these tests present, Jest prints `A worker process has failed to exit gracefully` at the end of a full-suite run. I investigated rather than ignoring it: it is reproducible (3/3 runs), disappears when both new test files are excluded, but does **not** reproduce when either file runs in isolation or under `--detectOpenHandles`, which forces serial execution. Every test passes and the run exits 0, so CI is unaffected.

I added mock-restore teardown in both files and cleared the React Query clients with `gcTime: 0`, which is correct hygiene regardless, but neither removed the warning. I did not chase it further since it changes no outcome — happy to dig in if you'd prefer it gone before merge.

Separately, I hit one pre-existing flaky test during bisection (a suite failed once, then passed on re-run with an identical command); it is unrelated to these changes.

## Note on the red `PR Governance` check

A timing race, not a real violation. `Enforce Minimum Label Usage` ran 4 seconds after the PR opened and failed with *"Please apply at least one meaningful label"*, but the labeling bot applied its labels moments later. The workflow does listen for `labeled`; however the bot labels using `GITHUB_TOKEN`, and GitHub suppresses workflow re-triggers from that token, so no re-run was ever queued.

The PR now carries `gssoc`, `gssoc:approved`, `type:feature`, `level:frontend`, `feat`, and `quality:exceptional`. Every other check is green, including `Unit Tests`, `Lint`, `Typecheck`, and `Verify Issue Assignment`.

### DIFF
--- a/frontend/src/__tests__/hooks/article/useGetArticleContent.test.ts
+++ b/frontend/src/__tests__/hooks/article/useGetArticleContent.test.ts
@@ -1,22 +1,47 @@
 import axios from 'axios';
 import {renderHook, waitFor} from '@testing-library/react-native';
 import {QueryClient, QueryClientProvider} from '@tanstack/react-query';
+import {configureStore} from '@reduxjs/toolkit';
+import {Provider} from 'react-redux';
 import React from 'react';
 import { useGetArticleContent } from '@/src/hooks/article/useGetArticleContent';
+import offlineReducer from '@/src/store/offlineSlice';
 
 jest.mock('axios');
 const mockedAxios = axios as jest.Mocked<typeof axios>;
 
+/** Query clients created during a test, cleared afterwards to free timers. */
+let clients: QueryClient[] = [];
+
+// The hook keeps the opened body in the session cache, so it needs a store.
 function makeWrapper() {
+  // gcTime 0: React Query's default five-minute garbage-collection timer
+  // outlives the test and keeps the Jest worker from exiting.
   const queryClient = new QueryClient({
-    defaultOptions: {queries: {retry: false}, mutations: {retry: false}},
+    defaultOptions: {
+      queries: {retry: false, gcTime: 0},
+      mutations: {retry: false},
+    },
   });
+  clients.push(queryClient);
+  const store = configureStore({reducer: {offline: offlineReducer}});
   return ({children}: {children: React.ReactNode}) =>
-    React.createElement(QueryClientProvider, {client: queryClient}, children);
+    React.createElement(Provider, {
+      store,
+      children: React.createElement(
+        QueryClientProvider,
+        {client: queryClient},
+        children,
+      ),
+    });
 }
 
 describe('useGetArticleContent', () => {
-  afterEach(() => jest.clearAllMocks());
+  afterEach(() => {
+    jest.clearAllMocks();
+    clients.forEach(client => client.clear());
+    clients = [];
+  });
 
   it('fetches article content successfully', async () => {
     mockedAxios.get.mockResolvedValueOnce({
--- a/frontend/src/__tests__/hooks/article/useGetArticleContent.test.tsx
+++ b/frontend/src/__tests__/hooks/article/useGetArticleContent.test.tsx
@@ -25,15 +25,11 @@ function makeWrapper() {
   });
   clients.push(queryClient);
   const store = configureStore({reducer: {offline: offlineReducer}});
-  return ({children}: {children: React.ReactNode}) =>
-    React.createElement(Provider, {
-      store,
-      children: React.createElement(
-        QueryClientProvider,
-        {client: queryClient},
-        children,
-      ),
-    });
+  return ({children}: {children: React.ReactNode}) => (
+    <Provider store={store}>
+      <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>
+    </Provider>
+  );
 }
 
 describe('useGetArticleContent', () => {
--- a/frontend/src/__tests__/hooks/article/useGetArticleDetail.test.ts
+++ b/frontend/src/__tests__/hooks/article/useGetArticleDetail.test.ts
@@ -1,22 +1,47 @@
 import axios from 'axios';
 import {renderHook, waitFor} from '@testing-library/react-native';
 import {QueryClient, QueryClientProvider} from '@tanstack/react-query';
+import {configureStore} from '@reduxjs/toolkit';
+import {Provider} from 'react-redux';
 import React from 'react';
 import {useGetArticleDetails} from '../../../hooks/article/useGetArticleDetail';
+import offlineReducer from '../../../store/offlineSlice';
 
 jest.mock('axios');
 const mockedAxios = axios as jest.Mocked<typeof axios>;
 
+/** Query clients created during a test, cleared afterwards to free timers. */
+let clients: QueryClient[] = [];
+
+// The hook keeps the opened article in the session cache, so it needs a store.
 function makeWrapper() {
+  // gcTime 0: React Query's default five-minute garbage-collection timer
+  // outlives the test and keeps the Jest worker from exiting.
   const queryClient = new QueryClient({
-    defaultOptions: {queries: {retry: false}, mutations: {retry: false}},
+    defaultOptions: {
+      queries: {retry: false, gcTime: 0},
+      mutations: {retry: false},
+    },
   });
+  clients.push(queryClient);
+  const store = configureStore({reducer: {offline: offlineReducer}});
   return ({children}: {children: React.ReactNode}) =>
-    React.createElement(QueryClientProvider, {client: queryClient}, children);
+    React.createElement(Provider, {
+      store,
+      children: React.createElement(
+        QueryClientProvider,
+        {client: queryClient},
+        children,
+      ),
+    });
 }
 
 describe('useGetArticleDetails', () => {
-  afterEach(() => jest.clearAllMocks());
+  afterEach(() => {
+    jest.clearAllMocks();
+    clients.forEach(client => client.clear());
+    clients = [];
+  });
 
   it('fetches article details successfully', async () => {
     mockedAxios.get.mockResolvedValueOnce({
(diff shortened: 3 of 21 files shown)


SB2318 (changes requested):
This will create one issue, A huge amount of storage on mobile device.

Please do one task: your system is really good, But The 20 article fetch and stored, There is a problem. Check on the details screen. When a read event status updating, If any reader fully read any article, then only stored it in device, Otherwise leave it.

Yeah, For Your implementation, don't change it , just clear the storage when app is unmounted.

SB2318 (commented):
@Anijesh  Your Flow should looks like,

* Fetch 20 Articles As per existing flow
* Store it on redux
        ↓
User Opens Article
        ↓
Read Event Tracking
        ↓
Fully Read?
  No -> Don't Store it Locally
 Yes ->   Only that article + 3-Day TTL
               
       ( Offline Access)

SB2318:
> @Anijesh Your Flow should looks like,
> 
> * Fetch 20 Articles As per existing flow
> * Store it on redux
>   ↓
>   User Opens Article
>   ↓
>   Read Event Tracking
>   ↓
>   Fully Read?
>   ↙          ↘
>   No           Yes
>   ↓             ↓
>   Don't Store   Store Locally
>   ↓
>   Max 20 + 7-Day TTL
>   ↓
>   Offline Access

Let me know if you not get any point.

SB2318:
@Anijesh any update?

Anijesh:
> @Anijesh any update?

Doing
It will complete by tomorrow

Anijesh:
@SB2318 Updated — reworked to the flow you drew. Pushed in `03013a0c`.

You were right about the storage: caching on open meant a copy of everything a reader glanced at. Storage now follows the read event.

| Step | Where it lives now |
|---|---|
| 20 articles fetched | Redux, existing flow — unchanged |
| User opens article | Redux session cache only, **nothing written to the device** |
| Read event / reached the end | Written to device: article + body, one entry |
| Not fully read | Never stored |
| Device cache | Max 20, 7-day TTL, expired entries pruned |
| App unmounts | Session cache dropped, device cache pruned |

## What changed

**`store/offlineSlice.ts`** (new) — session cache of opened articles and their bodies, capped at 20 and held in memory only. This is what makes an article stay readable if connectivity drops while the reader is still on it, which is the common case. It goes away with the app, so browsing costs no storage.

**`ArticleCacheUtils`** — `cacheArticleDetail`/`cacheArticleContent` are gone. One function now, `cacheFullyReadArticle`, which writes the article and its body together. `pruneArticleCache` drops expired entries when the app unmounts rather than waiting for the reader to browse again.

**`useSaveFullyReadArticle`** (new) — fires once per article from the fully-read handler on the detail screen, and skips anything not `published`, since a draft can still change or be discarded.

**Read hooks** — offline reads now try the session copy first, then an article the reader finished earlier. A 4xx is still never masked by cached content.

## Two calls I made, happy to change either

1. **The write happens when the reader reaches the end, not when the read-event POST is acknowledged.** Tying storage to the response means a dropped request silently costs the reader offline access — on exactly the flaky connection where they were going to need it. Same trigger point, just not waiting on the round trip.

2. **Guests are included.** They never report a read event but they do finish articles, so the save sits just outside the `!isGuest` check. Say the word if you want offline reading to be logged-in only.

## TTL

Your review said 3 days and the flow comment after it said 7, so I kept 7 to match the acceptance criteria on #1880. One constant (`CACHE_TTL_DAYS`) if you want it at 3.

## Tests

**143 suites, 537 tests, all passing.** 34 across the offline files:

- `offlineSlice.test.ts` (5) — session cache bounded at 20, no duplicates, cleared on unmount.
- `ArticleCacheUtils.test.ts` (10) — eviction past 20, re-read survives eviction, 7-day expiry, body kept across a re-read, prune on unmount, corrupt-data recovery.
- `useSaveFullyReadArticle.test.ts` (5) — stores on finish, **writes once however many bottom-of-page scroll events arrive**, skips unpublished, does nothing while loading.
- `useOfflineArticleReading.test.ts` (8) — **opening an article writes nothing to the device**, session copy served when connectivity drops, fully read article served in a later session, still fails for an article never read, 404 not masked.

`tsc --noEmit` and `eslint` clean on the touched files.

## The Jest worker warning from my last comment

Found it. The two article hook suites were leaking React Query's default five-minute garbage-collection timer, so the worker outlived the run and had to be force exited. `gcTime: 0` plus clearing the clients in `afterEach` fixes those suites — that pair is now silent.

The warning still shows on a full-suite run from somewhere else in the repo: it reproduces with all six offline test files excluded, so it is not coming from this PR. Same for `useGetWeeklyWellness.test.ts`, which fails intermittently and passes on re-run; it doesn't touch any of this code. Both look like pre-existing issues worth a separate look — I can open an issue if you'd like.

Anijesh:
One more pass over this while CI ran, and I found a performance problem worth fixing in the same area — pushed as `699d3d8d`.

The "Available Offline" badge asked storage per card. That answer lives in a single AsyncStorage entry which now holds the cached articles *with their bodies*, so a list of twenty cards read and parsed that whole payload twenty times to reach the same answer — on the JS thread, on the home screen. The lookup goes through React Query now, so one read serves every card on screen. A test pins it: four cards asking produces exactly one read of the cache.

Full suite after all three commits: **144 suites, 541 tests, passing**. Lint and typecheck green, and all 12 checks passed on the previous commit.

SB2318 (approved):
Great! It will be a count as a major contribution

Issue #2348, opened 2026-08-10
🐞[Bug]: Improve unclear Go to Top button icon and accessibility


aayan-rashid:
### Related Area

Frontend

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

The Go to Top control uses an isolated upward-arrow icon without a clear label or visual context. It may not be immediately obvious to users that the button scrolls the page back to the top.

Steps to Reproduce

Open the website.
Scroll down the page.
Locate the floating/navigation control containing the upward arrow.
Observe the button.

### Expected behavior

The control should clearly communicate that it scrolls the user back to the top of the page.

For example, it could use:

A recognizable ↑ icon with a tooltip such as "Back to top", or
A clearly labeled "Back to top" button.

It should also have an appropriate accessible label for screen readers.

### Add ScreenShots

<img width="165" height="141" alt="Image" src="https://github.com/user-attachments/assets/dde745e2-1ba5-4782-9f2e-d04950b826ee" />

### What browsers are you seeing the problem on?

Chrome

### Record

- [x] I have read the Contributing Guidelines
- [ ] I'm a GSSOC'24 contributor
- [x] I'm a GSSOC'26 contributor
- [ ] I'm a IEEE IGDTUW contributor
- [x] I want to work on this issue

PR #2375, opened 2026-08-13
fix: improve Go to Top button accessibility and add tooltip

gauridhingra03:
## Description
Fixes the unclear "Go to Top" button by adding a tooltip and proper accessibility label, so users and screen reader users can clearly understand what the button does.

## Changes
- Added `title="Back to top"` on the button for a hover tooltip
- Updated `aria-label` to `"Scroll back to top"` for clearer screen reader announcement
- Added `aria-hidden="true"` on the decorative SVG and arrow span so screen readers don't double-announce the icon content

## Testing
- Verified tooltip appears on hover (native browser tooltip)
- Verified `aria-label` and `title` attributes render correctly via DevTools inspection
- Verified `aria-hidden="true"` applied to SVG and arrow span
- Confirmed scroll-to-top click functionality still works as expected

Fixes #2348

### DIFF
--- a/web/src/components/ScrollToTop.tsx
+++ b/web/src/components/ScrollToTop.tsx
@@ -41,9 +41,10 @@ export default function ScrollToTop() {
     <button
       className="scroll-to-top"
       onClick={scrollToTop}
-      aria-label="Scroll to top"
+      aria-label="Scroll back to top"
+      title="Back to top"
     >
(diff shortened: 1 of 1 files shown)


gauridhingra03:
Hi, could you please add the gssoc:approved label along with the merge when you review this? Thanks!

SB2318 (approved):

PR #2391, opened 2026-08-14
Fix #1725: Chat typing bubble shows on load

### LINKED ISSUE #1725: Fix Multiple UI State and Interaction Bugs Across Home Feed, Chatbot, Social Follow, and Genre Selection
Several user-facing issues have been identified across the application that negatively impact usability and functionality. These include incorrect home feed state rendering, chatbot typing indicator behavior, social follow notification inconsistencies, and genre/tag selection problems.

Problems Identified
1. Genre/Tag Deselection Not Working

Users are unable to deselect previously selected genres/tags because the selection logic relies on strict object reference comparison rather than unique identifiers.

Expected Behavior

Users should be able to select and deselect genres/tags reliably.
Selection state should be determined using unique identifiers such as id, _id, or name.
2. Social Follow Notification Race Condition

Follow notifications may be emitted with an incorrect or empty user ID due to asynchronous React state updates.

Expected Behavior

Follow notifications should always be sent with the correct follower ID.
Notification emission should not depend on potentially stale state values.
3. Chatbot Typing Indicator Incorrect on Initial Load

The chatbot typing indicator appears active immediately after the screen loads and may remain visible unnecessarily.

Expected Behavior

Typing indicator should remain hidden on initial screen load.
It should only appear while a chatbot response is actively being generated.
4. Home Feed Loading and Offline States Swapped

The Home Feed currently displays incorrect UI states:

Loading state may appear when the device is offline.
Offline state may appear during data loading.

Additionally, empty article responses may result in a blank screen.

Expected Behavior

Loading state should appear only while content is being fetched.
Offline state should appear when network connectivity is unavailable.
Empty article responses should render an appropriate empty-state screen.
Proposed Solution
Replace object reference checks for genre selection with identifier-based matching.
Pass the follower ID directly when emitting follow notifications.
Manage chatbot typing indicator state around API request lifecycle events.
Correct Home Feed state rendering logic and introduce proper empty-state handling.
Acceptance Criteria

Genres/tags can be selected and deselected correctly.

Follow notifications always contain the correct follower ID.

Chatbot typing indicator only appears during active response generation.

Home Feed displays the correct loading and offline states.

Empty article responses render a dedicated empty-state screen.

Unit tests cover all affected components and pass successfully.

saurabhhhcodes:
## Description

Fixes the chatbot typing indicator bug from issue #1725 (part 3: Chatbot Typing Indicator Incorrect on Initial Load).

`ChatbotScreen` initialized `isTyping` to `true`, so the "typing..." bubble rendered as soon as the screen loaded, before any message had been sent.

## Changes

- `frontend/src/screens/ChatbotScreen.tsx`:
  - Initialize `isTyping` to `false` so the indicator stays hidden on screen load.
  - Added an effect that syncs `isTyping` with the mutation's `isPending` state, so the typing indicator appears only while the AI response is being generated and is cleared when the request settles (success, error, or cancellation).
  - Removed the now-redundant `safeSetIsTyping(false)` from the conversation-loading effect (initial state is already `false`; this also avoids clearing the indicator if a request is in flight when history finishes loading).
- `frontend/src/screens/__tests__/ChatbotScreen.test.tsx` (new): unit tests asserting the typing indicator is hidden on initial load and only shown while a response is pending.

## Verification

- `tsc --noEmit` in `frontend/`: 58 pre-existing errors before and after — no new errors added.
- `npx jest src/screens/__tests__/ChatbotScreen.test.tsx`: 2 tests pass.

Fixes #1725

### DIFF
--- a/frontend/src/screens/ChatbotScreen.tsx
+++ b/frontend/src/screens/ChatbotScreen.tsx
@@ -53,7 +53,7 @@ const ChatbotScreen = ({navigation}: ChatBotScreenProps) => {
   const {isConnected} = useSelector((state: any) => state.network);
 
   const [messages, setMessages] = useState<IMessage[]>([]);
-  const [isTyping, setIsTyping] = useState<boolean>(true);
+  const [isTyping, setIsTyping] = useState<boolean>(false);
   const isMountedRef = useRef(true);
   const dispatch = useDispatch();
   const {data: user} = useGetProfile();
@@ -103,10 +103,12 @@ const ChatbotScreen = ({navigation}: ChatBotScreenProps) => {
         },
         ...refined.reverse(),
       ]);
-
-      safeSetIsTyping(false);
     }
   }, [conversations, safeSetIsTyping, safeSetMessages]);
+
+  useEffect(() => {
+    safeSetIsTyping(messageProcessPending);
+  }, [messageProcessPending, safeSetIsTyping]);
 
   const convertToGiftedFormat = (items: Message[]): IMessage[] => {
     return items.map(m => ({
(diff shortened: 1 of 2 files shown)


SB2318:
@saurabhhhcodes Please resolve the merge conflict.

image

PR #2409, opened 2026-08-14
Fix : Scope `zod-validation-error` override to `eslint-plugin-react-hooks`

### LINKED ISSUE #2373: Bug: `npx eslint` crashes repo-wide with `ERR_PACKAGE_PATH_NOT_EXPORTED`
## Summary

Running ESLint inside the `frontend/` directory currently fails before any project files are linted.

The error is related to `zod-validation-error` and its `./v4` package export:

```text
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]:
Package subpath './v4' is not defined by "exports"
in .../node_modules/zod-validation-error/package.json
```

Because ESLint crashes during startup, developers currently cannot get any useful lint output from the frontend.

The issue has been reproduced on both **Linux and Windows**, so it does not appear to be specific to a particular operating system.

## Steps to Reproduce

1. Open the `frontend` directory.
2. Install the dependencies.
3. Run ESLint:

```bash
cd frontend
npm install
npx eslint . --ext .ts,.tsx
```

## Actual Behavior

ESLint crashes immediately with:

```text
Oops! Something went wrong! :(

ESLint: 9.39.5

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]:
Package subpath './v4' is not defined by "exports"
in .../node_modules/zod-validation-error/package.json
```

No project files are linted.

## Expected Behavior

ESLint should start normally and report any lint errors or warnings found in the project.

## Possible Root Cause

This looks like a dependency version mismatch somewhere in the ESLint configuration or its dependency chain.

Something appears to be trying to import the `./v4` subpath from `zod-validation-error`, while the installed version of `zod-validation-error` does not expose that subpath through its `exports` field.

Some possible causes are:

* `zod-validation-error` was updated or downgraded without updating a dependent package.
* An ESLint plugin or shared configuration expects a different version.
* npm resolved incompatible versions in the dependency tree.
* Dependencies were installed using `--legacy-peer-deps`, allowing an incompatible combination of packages.

The exact package causing the import should be identified before changing or pinning any dependency version.

## Suggested Investigation

First, inspect the installed versions:

```bash
npm ls zod-validation-error zod
```

Then:

1. Identify which package depends on `zod-validation-error`.
2. Check which version of `zod-validation-error` that package expects.
3. Identify which package is attempting to import the `./v4` subpath.
4. Check the relevant package versions and changelogs for compatibility information.
5. Update, downgrade, or pin the dependency only after identifying the actual incompatibility.
6. Run ESLint again:

```bash
npx eslint .
```

7. Confirm that ESLint starts successfully and produces actual lint output.
8. Handle any project lint errors separately after the startup issue is fixed.

The goal should be to fix the underlying dependency mismatch rather than forcing an arbitrary version of `zod-validation-error.

## Why This Needs Investigation

This is more than a normal linting error. ESLint is failing during startup, which means developers currently cannot use the frontend's linting setup at all.

The dependency chain needs to be traced to understand why a package is requesting an export that the installed version does not provide.

## Validation

After applying the fix, I will verify that:

```bash
npm install
npx eslint .
```

runs without the `ERR_PACKAGE_PATH_NOT_EXPORTED` startup error and that ESLint is able to report actual lint results from the project.

## Suggested Labels

`bug` `tooling` `level2`

bh462007:
**Fixes #2373**

## Summary

ESLint was crashing during startup with `ERR_PACKAGE_PATH_NOT_EXPORTED` before it could lint any files.

This PR fixes the dependency mismatch by adding a scoped npm override for `zod-validation-error` under `eslint-plugin-react-hooks`.

## Root Cause

`eslint-plugin-react-hooks@7.1.1`, which is pulled in through `eslint-config-expo`, supports both `zod-validation-error` 3.x and 4.x:

```text
zod-validation-error: ^3.5.0 || ^4.0.0
```

The project uses `zod@4.4.3`, so `eslint-plugin-react-hooks` attempts to load the `./v4` subpath from `zod-validation-error`.

However, npm resolved `zod-validation-error` to `3.5.4`.

Although `3.5.4` satisfies the `^3.5.0` part of the dependency range, it does not expose the `./v4` subpath. That export was introduced in `zod-validation-error@4.0.0`.

As a result, `eslint-plugin-react-hooks` attempts to load:

```text
zod-validation-error/v4
```

and Node.js throws:

```text
ERR_PACKAGE_PATH_NOT_EXPORTED
```

There is also a separate dependency chain through `eslint-plugin-react-compiler@19.1.0-rc.2`, which depends on `zod-validation-error@^3.0.3` and correctly uses its `3.5.4` copy. That dependency is unrelated to this issue and is intentionally left unchanged.

## Fix

Added a scoped override to `frontend/package.json`:

```json
"overrides": {
  "eslint-plugin-react-hooks": {
    "zod-validation-error": "^4.0.2"
  }
}
```

This ensures that `eslint-plugin-react-hooks` gets a compatible `zod-validation-error` version that provides the required `./v4` export.

The override is scoped specifically to `eslint-plugin-react-hooks`, so the `eslint-plugin-react-compiler` dependency remains on `zod-validation-error@3.5.4`.

### Dependency tree

Ran:

```bash
npm ls zod-validation-error
```

Confirmed:

* `eslint-plugin-react-hooks` → `zod-validation-error@4.0.2`
* `eslint-plugin-react-compiler` → `zod-validation-error@3.5.4`
* No `invalid` or `ELSPROBLEMS` warnings

### ESLint

Ran:

```bash
npx eslint . --ext .ts,.tsx
```

ESLint now starts successfully and reports the project's actual lint warnings instead of crashing.

The verification produced approximately 158–163 warnings with **0 startup crashes**.

### TypeScript

Ran:

```bash
npx tsc --noEmit
```

The error count remained **1117 before and after** the change.

The comparison was performed on the same machine using `git stash` / `git stash pop`, confirming that the dependency override did not introduce additional TypeScript errors.

## Note

The project currently requires `--legacy-peer-deps` during installation because of an existing peer dependency conflict between `lottie-react-native` and `@lottiefiles/dotlottie-react`.

This is unrelated to the current fix and is not caused by the `zod-validation-error` override.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -171,5 +171,10 @@
       }
     }
   },
-  "packageManager": "yarn@4.18.0"
+  "packageManager": "yarn@4.18.0",
+  "overrides": {
+    "eslint-plugin-react-hooks": {
+      "zod-validation-error": "^4.0.2"
+    }
+  }
 }
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -171,10 +171,10 @@
       }
     }
   },
-  "packageManager": "yarn@4.18.0",
   "overrides": {
     "eslint-plugin-react-hooks": {
       "zod-validation-error": "^4.0.2"
     }
-  }
+  },
+  "packageManager": "yarn@4.18.0"
 }


SB2318 (changes requested):
Don't specify Package manager here, because currently we are at `yarn 4.21`

bh462007:
Reordered it so overrides now comes before packageManager. The diff is cleaner now and doesn’t touch the packageManager line at all — it just adds the new override block above it.

I’ve pushed this as a new commit. Let me know if this looks good!

SB2318 (approved):

Issue #1761, opened 2026-06-20
🐞[Bug]:,Medical Glossary Search Matches Definitions Instead of Glossary Term Names


Harsh2865:
### Related Area

Frontend

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

The Medical Glossary search functionality currently matches search keywords against glossary definitions/content instead of primarily matching glossary term names.

For example, when searching for "Ast", the expected result is "Asthma". However, the search also returns unrelated glossary entries because the keyword may appear somewhere within the definition text.

Similarly, searching for "Press" returns both "Blood Pressure" and "Hypertension". While "Blood Pressure" is relevant, "Hypertension" does not contain the searched keyword in its title and appears only because the keyword is found within related content or descriptions.

This behavior reduces search precision and creates a confusing user experience, especially for users who expect glossary search to behave like a dictionary where results are based on term names.

### Expected behavior

The glossary search should primarily match glossary term names.

Examples:
- Searching "Ast" should return Asthma.
- Searching "Press" should return Blood Pressure.
- Searching "Hyper" should return Hypertension.

Results should not be displayed solely because the search term appears somewhere inside the glossary definition.

If searching within definitions is desired, it should be implemented as a separate advanced search feature rather than the default glossary search behavior.

### Add ScreenShots

<img width="1917" height="990" alt="Image" src="https://github.com/user-attachments/assets/bab4c11f-86f7-4e11-b10e-00d9c336a504" />

Screenshot 1:

Search query: "Ast"

Expected:
Results should prioritize glossary terms whose names contain the searched keyword.

Actual:
The search system searches across glossary content and definitions instead of focusing on glossary term names. This can return additional matches that are not directly related to the glossary title being searched.

<img width="1902" height="1025" alt="Image" src="https://github.com/user-attachments/assets/1f5c7d7c-4444-440a-b6eb-17dbeec70bad" />

Screenshot 2:
Search query: "Press"

Expected:
Only "Blood Pressure" should be displayed because its glossary term contains the searched keyword.

Actual:
Both "Blood Pressure" and "Hypertension" are displayed. "Hypertension" does not contain the keyword "Press" in its title and appears because the search is also checking glossary descriptions/content.

This demonstrates that the current search logic is not prioritizing glossary term names, resulting in less accurate search results.

Optional Enhancement

If searching within glossary definitions is an intended feature, consider introducing an Advanced Search mode.

[...]

PR #2414, opened 2026-08-15
fix(web): glossary search matches term names only

singhanurag0317-bit:
## Summary
Fixes #1761 — the Medical Glossary search now matches query terms against glossary **term names** only, instead of matching against the full definition text. This makes search behave like a dictionary lookup and removes irrelevant results that matched only inside definitions.

## Problem
`MedicalGlossaryExplorer.tsx` filtered entries with:
```ts
entry.term.toLowerCase().includes(q) || entry.definition.toLowerCase().includes(q)
```
So searching "Press" returned both "Blood Pressure" (relevant) and "Hypertension" (matched only because the keyword appears inside its definition/related content).

## Change
```ts
const matchesSearch = entry.term.toLowerCase().includes(q);
```

## Behavior after fix
- "Ast" → Asthma (plus any other term whose **name** contains "ast")
- "Press" → Blood Pressure only
- "Hyper" → Hypertension
- Empty query still shows all terms; category filter still works; related-term chips still jump to the term name.

Per the issue, definition-content search is intentionally not part of the default glossary search (could be a separate advanced feature later).

## Verification
- No `node_modules` available locally, so `npm test`/`lint` couldn't run in `web/`; the change is a single boolean condition and is covered by CI.

### DIFF
--- a/web/src/app/[locale]/medical-glossary/MedicalGlossaryExplorer.tsx
+++ b/web/src/app/[locale]/medical-glossary/MedicalGlossaryExplorer.tsx
@@ -86,9 +86,7 @@ export default function MedicalGlossaryExplorer() {
   const filteredEntries = useMemo(() => {
     const q = searchQuery.toLowerCase();
     return glossaryEntries.filter((entry) => {
-      const matchesSearch =
-        entry.term.toLowerCase().includes(q) ||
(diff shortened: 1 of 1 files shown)


SB2318 (approved):
Bro, GSSOC Running or not?

